PR #4, opened 2017-05-05
add cake build

sphiecoh:


build.cake
```diff
@@ -0,0 +1,41 @@
...
+    .Does(() =>
+{
+    var settings = new DotNetCorePackSettings
+    {
+        Configuration = "Release",
+        OutputDirectory = buildArtifacts,
+        NoBuild = true,
+        VersionSuffix = versionSuffix
+    };
+    var projects = GetFiles("./src/**/*.csproj");
```
jchannon: That should probably hardcode the path  and just call `DotNetCorePack("./src/Botwin", settings);`

build/version.props
```diff
@@ -0,0 +1,7 @@
+<Project>
+ <PropertyGroup>
+    <VersionPrefix>1.0.0</VersionPrefix>
+    <VersionSuffix>beta2</VersionSuffix>
+    <VersionSuffix Condition=" '$(VersionSuffix)' != '' AND '$(BuildNumber)' != '' ">$(VersionSuffix)-$(BuildNumber)</VersionSuffix>
```
jchannon: Where does BuildNumber come from?

build.sh
```diff
@@ -0,0 +1,131 @@
+#!/usr/bin/env bash
```
jchannon: guessing this is a copy/paste?

build.ps1
```diff
@@ -0,0 +1,228 @@
+##########################################################################
```
jchannon: guessing this is a copy/paste?

build.sh
```diff
@@ -0,0 +1,131 @@
...
+    fi
+fi
+
+# Restore tools from NuGet.
+pushd "$TOOLS_DIR" >/dev/null
+if [ ! -f "$PACKAGES_CONFIG_MD5" ] || [ "$( cat "$PACKAGES_CONFIG_MD5" | sed 's/\r$//' )" != "$( $MD5_EXE "$PACKAGES_CONFIG" | awk '{ print $1 }' )" ]; then
+    find . -type d ! -name . | xargs rm -rf
+fi
+
+mono "$NUGET_EXE" install -ExcludeVersion
```
jchannon: Do we need mono? Ideally not? @devlead

build.sh
```diff
@@ -0,0 +1,131 @@
...
+
+# Make sure that Cake has been installed.
+if [ ! -f "$CAKE_EXE" ]; then
+    echo "Could not find Cake.exe at '$CAKE_EXE'."
+    exit 1
+fi
+
+# Start Cake
+if $SHOW_VERSION; then
+    exec mono "$CAKE_EXE" -version
```
jchannon: Can cake run without mono? @devlead

build.sh
Sphiecoh: default cake bootstrapper

build.ps1
Sphiecoh: default cake bootstrapper

build.sh
devlead: Yes It's possible to run without mono, you'll have to fetch cake using curl&unzip something like
This example: https://github.com/devlead/BitbucketPipelinesShield/blob/master/build.sh

Sphiecoh:
I have removed mono

jchannon:
I had a play and have it working with codeship currently.  Will keep this
PR open for a bit just to see how we get on with codeship first. hope thats
ok?

PR #11, opened 2017-06-21
add Patch and OPTIONS to botwin module

Sphiecoh:


jchannon:
Always a pleasure ;)

Thanks!

Issue #13, opened 2017-08-19
Don't build the service provider to detect if a service is registered


davidfowl:
https://github.com/jchannon/Botwin/blob/92bb7732be433cb70c22fe76e694a793d246db5b/src/BotwinExtensions.cs#L130

Instead, you should use TryAdd*. In this case `TryAddSingleton<IAssemblyProvider, AssemblyProvider>()`. See https://github.com/aspnet/Mvc/blob/900a5c7c4c789e71bf18e8e90bf9b28b4e51013f/src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreServiceCollectionExtensions.cs#L145

jchannon:
Cheers will amend that.

On another note, is it ok to build it to get the provider to "get" a
registered item?

PR #14, opened 2017-09-12
Update to netstandard2.0

petedishman:
Hi,

Randomly saw your tweet about updating to netstandard2.0 and had been looking for something to get a bit of a look in to dotnetcore, so had a go.

Seems to work alright to me, not sure what VS2017 has done to the .sln though.
Also made what I believe is the change suggested by David Fowler in #13 as without changes it was blowing up due to a null assemblyProvider.

The mediaType check in DefaultJsonResponseNegotiator could probably be more optimal but I didn't see anything after a quick look.

Hope it helps

### DIFF
--- a/src/Botwin.csproj
+++ b/src/Botwin.csproj
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <TargetFramework>netstandard2.0</TargetFramework>
     <VersionPrefix>1.0.1</VersionPrefix>
     <Authors>Jonathan Channon</Authors>
     <Description>Botwin is a library that allows Nancy-esque routing for use with ASP.Net Core.</Description>
@@ -10,9 +10,9 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="FluentValidation" Version="6.4.1" />
-    <PackageReference Include="Microsoft.AspNetCore.Routing" Version="2.0.0-preview1-final" />
+    <PackageReference Include="Microsoft.AspNetCore.Routing" Version="2.0.0" />
     <PackageReference Include="Newtonsoft.Json" Version="10.0.2" />
     <PackageReference Include="System.ValueTuple" Version="4.3.0" />
-    <PackageReference Include="microsoft.extensions.dependencyinjection" Version="2.0.0-preview1-final" />
+    <PackageReference Include="microsoft.extensions.dependencyinjection" Version="2.0.0" />
   </ItemGroup>
 </Project>
--- a/src/Botwin.csproj
+++ b/src/Botwin.csproj
@@ -13,6 +13,6 @@
     <PackageReference Include="Microsoft.AspNetCore.Routing" Version="2.0.0" />
     <PackageReference Include="Newtonsoft.Json" Version="10.0.2" />
     <PackageReference Include="System.ValueTuple" Version="4.3.0" />
-    <PackageReference Include="microsoft.extensions.dependencyinjection" Version="2.0.0" />
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="2.0.0" />
   </ItemGroup>
 </Project>
--- a/src/BotwinExtensions.cs
+++ b/src/BotwinExtensions.cs
@@ -9,6 +9,7 @@ namespace Botwin
     using Microsoft.AspNetCore.Http;
     using Microsoft.AspNetCore.Routing;
     using Microsoft.Extensions.DependencyInjection;
+    using Microsoft.Extensions.DependencyInjection.Extensions;
 
     public static class BotwinExtensions
     {
@@ -127,16 +128,12 @@ namespace Botwin
         public static void AddBotwin(this IServiceCollection services)
         {
             //Get IAssemblyProvider, if not found register default provider.
+            services.TryAddSingleton<IAssemblyProvider, AssemblyProvider>();
+
             var provider = services.BuildServiceProvider();
             var assemblyProvider = provider.GetService<IAssemblyProvider>();
-            if (assemblyProvider == null)
-            {
-                services.AddSingleton<IAssemblyProvider, AssemblyProvider>();
-            }
-            assemblyProvider = provider.GetService<IAssemblyProvider>();
 
             services.AddRouting();
-
 
             var modules = assemblyProvider.GetAssembly().GetTypes().Where(t => typeof(BotwinModule).IsAssignableFrom(t) && t != typeof(BotwinModule));
             foreach (var module in modules)
--- a/src/DefaultJsonResponseNegotiator.cs
+++ b/src/DefaultJsonResponseNegotiator.cs
@@ -12,7 +12,7 @@
     {
         public bool CanHandle(IList<MediaTypeHeaderValue> accept)
         {
-            return accept.Any(x => x.MediaType.IndexOf("json", StringComparison.OrdinalIgnoreCase) >= 0);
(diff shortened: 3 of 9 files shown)


src/DefaultJsonResponseNegotiator.cs
```diff
@@ -12,7 +12,7 @@ public class DefaultJsonResponseNegotiator : IResponseNegotiator
     {
         public bool CanHandle(IList<MediaTypeHeaderValue> accept)
         {
-            return accept.Any(x => x.MediaType.IndexOf("json", StringComparison.OrdinalIgnoreCase) >= 0);
+            return accept.Any(x => x.MediaType.ToString().IndexOf("json", StringComparison.OrdinalIgnoreCase) >= 0);
```
jchannon: Don't think you need the ToString as it's implicit

test/ResponseNegotiatorTests.cs
```diff
@@ -61,7 +61,7 @@ internal class TestResponseNegotiator : IResponseNegotiator
     {
         public bool CanHandle(IList<MediaTypeHeaderValue> accept)
         {
-            return accept.Any(x => x.MediaType.IndexOf("foo/bar", StringComparison.OrdinalIgnoreCase) >= 0);
+            return accept.Any(x => x.MediaType.ToString().IndexOf("foo/bar", StringComparison.OrdinalIgnoreCase) >= 0);
```
jchannon: Don't think you need the ToString as it's implicit

src/DefaultJsonResponseNegotiator.cs
slang25: It is needed, or `.Value`

petedishman:
yeah, MediaType is a StringSegment so it's IndexOf only takes Char's not
String's

slang25:
Cool, this looks good, I did the same thing: https://github.com/jchannon/Botwin/compare/master...slang25:aspnetcore2?expand=1

I think you need to update the `Dockerfile` and you should have a green build.

I also had a go at resolving David Fowler's comment, but got a bit stuck. I presume the existing logic is so that if there is already a registered service, you use that. It then gets tricky cause how do you register additional services without building and retrieving the current implementation? I thought his concern was with calling `BuildServiceProvider()`. @davidfowl can  you clarify?

slang25:
You could just do for example:
```csharp
var assemblyProvider = new AssemblyProvider();
services.TryAddSingleton<IAssemblyProvider>(assemblyProvider);            
```
If you don't care about using an existing definition.

jchannon:
Were the tests giving false positives then because we didn't have the `.Value` / `ToString`?

jchannon:
@slang25 correct. If the user has already registered use that else add a default and use that

petedishman:
wouldn't compile without the .ToString()
I'm on the wrong computer to check at the moment, but presume it was a
change introduced when upgrading to the 2.0.0 final packages

jchannon:
Ah, makes sense

jchannon:
The CI stuff is borked. I think its a CodeShip issue

slang25:
Just to be clear, the `TryAddSingleton` is an improvement and retains the current behaviour. I'm just wondering if there is a "right way" that @davidfowl had in mind which can actually close #13, it's still a mystery to me.

jchannon:
I assume TryAdd will fail if the user already registered one however I dont there is another approach to get something out if you want it during registration.  Seems hacky but understandable

slang25:
@jchannon Yeah that is my understanding, it will be like "Oh there's already one there, that's cool", then later if you want to get that concrete thing you'll have to build the provider and grab it.

Another approach (even more hacky IMO) could be to have `AssemblyProvider` take an `IServiceCollection` and add the services there, but that's not better.

jchannon:
I think this good to go tbh in terms of upgrading to netstandard. Thoughts?

slang25:
Yeah it's great. Just a minor change to the Dockerfile is needed, but that could be done when you looks at getting CodeShip working: https://github.com/petedishman/Botwin/blob/master/Dockerfile#L1

src/Botwin.csproj
```diff
@@ -10,9 +10,9 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="FluentValidation" Version="6.4.1" />
-    <PackageReference Include="Microsoft.AspNetCore.Routing" Version="2.0.0-preview1-final" />
+    <PackageReference Include="Microsoft.AspNetCore.Routing" Version="2.0.0" />
     <PackageReference Include="Newtonsoft.Json" Version="10.0.2" />
     <PackageReference Include="System.ValueTuple" Version="4.3.0" />
-    <PackageReference Include="microsoft.extensions.dependencyinjection" Version="2.0.0-preview1-final" />
+    <PackageReference Include="microsoft.extensions.dependencyinjection" Version="2.0.0" />
```
davidfowl: Fix the casing

src/BotwinExtensions.cs
```diff
@@ -127,17 +128,13 @@ private static void ApplyGlobalBeforeHook(IApplicationBuilder builder, BotwinOpt
         public static void AddBotwin(this IServiceCollection services)
         {
             //Get IAssemblyProvider, if not found register default provider.
+            services.TryAddSingleton<IAssemblyProvider, AssemblyProvider>();
+
             var provider = services.BuildServiceProvider();
```
davidfowl: This is pretty gross still...

jchannon:
actually might as well do it here. @petedishman could remove the `preview` string in the dockerfile please?

src/BotwinExtensions.cs
jchannon: Alternative?

davidfowl: Maybe just change this entire pattern. Remove the assembly provider since it's too early to discover services. Make it a `Func` instead of an interface?

jchannon: And resolve the Func here? https://github.com/jchannon/Botwin/blob/master/src/BotwinExtensions.cs#L29

davidfowl: Here https://github.com/jchannon/Botwin/blob/master/src/BotwinExtensions.cs#L136.

I'm saying to avoid adding things in the service provider that affect adding more services.

slang25: Right, so I'm taking that as you wouldn't touch the `IServicesCollection` with `IAssemblyProvider`, you'd have some overloads to `AddBotwin` that take optional `Assemby`s/`Type`s/`IAssemblyProvider`s etc...
Just like AutoMapper does here: https://github.com/AutoMapper/AutoMapper.Extensions.Microsoft.DependencyInjection/blob/master/src/AutoMapper.Extensions.Microsoft.DependencyInjection/ServiceCollectionExtensions.cs#L37

jchannon: I had a play with that approach last night and it seemed to work well. The AddBotwin takes in params Assembly and we use it at that point to find modules etc but I also register IEnumerable<Assembly> so that they can be pulled out in places like this : https://github.com/jchannon/Botwin/blob/master/src/BindExtensions.cs#L25

davidfowl: You scan all types in the assembly potentially per request or am I missing something?

jchannon: Ah good point. Maybe best to do that check in the AddBotwin method and register the `IEnumerable< AbstractValidator<>>` and then pull them out on the Bind method

jchannon: I think I will merge this as is then put in the fixes I spiked last night on another commit that removes IAssemblyProvider.  Will cc people on that PR when I do it

jchannon:
Thanks @petedishman awesome stuff!! ❤️  👍

PR #22, opened 2017-10-06
Added .travis.yml

JoeStead:
Added a travis file to test everything in the test directory and pack everything in the src directory.

It will need some tweaking to actually auto deploy new versions of packages, but as a first pass this will at least work for testing PRs.

You will need to do a bit of setup on the Travis CI side of things (by simply checking a few boxes) after authorising Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,7 @@
+language: csharp
+mono: none
+dotnet: 2.0.0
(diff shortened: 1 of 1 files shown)


JoeStead:
You can see it working on my fork here: https://travis-ci.org/JoeStead/Botwin

PR #23, opened 2017-10-06
Added sensible extensions for dealing with query strings

JoeStead:
Added extension methods for working with query strings.

### DIFF
--- a/test/BotwinModuleTests.cs
+++ b/test/BotwinModuleTests.cs
@@ -174,5 +174,28 @@
             var body = await response.Content.ReadAsStringAsync();
             Assert.True(body.Contains("After"));
         }
+        
+        [Fact]
+        public async Task Should_return_GET_requests_with_parsed_querystring()
+        {
+            const int idToTest = 69; 
+            var response = await this.httpClient.GetAsync($"/querystring?id={idToTest}");
+
+            var body = await response.Content.ReadAsStringAsync();
+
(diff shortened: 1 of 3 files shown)


test/BotwinModuleTests.cs
```diff
@@ -174,5 +174,28 @@ public BotwinModuleTests()
...
+            var response = await this.httpClient.GetAsync($"/querystring?id={idToTest}");
+
+            var body = await response.Content.ReadAsStringAsync();
+
+            Assert.Equal(200, (int)response.StatusCode);
+            Assert.True(body.Contains($"Managed to parse an int {idToTest}"));
+        }
+        
+        [Fact]
+        public async Task Should_return_GET_requests_with_multiple_parsed_querystring()
```
jchannon: Worth a test if you do `?id=1,2,3`?

JoeStead: Not tested, but I imagine it would fail on the parsing, because the query collection would return that as one value.

jchannon: not sure it would because query[id] returns ienumerable<string> so i'd expect that with 1,2,3 in it

JoeStead: That returns one value. I don't think we should go down the route of splitting that manually either

JoeStead: Thinking about it, another overload to handle that would be quite nice.

PR #24, opened 2017-10-07
Handle querystring multiple and default value

jchannon:


JoeStead:
I was thinking about this earlier. This is a bad idea. Some cultures will use the comma for a decimal point, which will screw up anyone who wants to use floats/double/decimal in a query string. May want to rethink this

jchannon:
The comma comes from MS. Query returns the collection with a string.Join.
Decompile Query and it shows it

PR #26, opened 2017-10-07
WIP initial local function spike

jchannon:
need to double check the async/await lack of

### DIFF
--- a/src/BotwinExtensions.cs
+++ b/src/BotwinExtensions.cs
@@ -4,6 +4,7 @@ namespace Botwin
     using System.IO;
     using System.Linq;
     using System.Reflection;
+    using System.Threading.Tasks;
     using FluentValidation;
     using Microsoft.AspNetCore.Builder;
     using Microsoft.AspNetCore.Http;
@@ -40,7 +41,7 @@ namespace Botwin
                     handler = CreateModuleBeforeAfterHandler(module, route);
 
                     var finalHandler = CreateFinalHandler(handler, statusCodeHandlers);
-                    
+
                     routeBuilder.MapVerb(route.verb, route.path, finalHandler);
                 }
             }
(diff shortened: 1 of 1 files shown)


jchannon:
After reading a bit of this not sure we save anything in terms of perf https://blogs.msdn.microsoft.com/seteplia/2017/10/03/dissecting-the-local-functions-in-c-7/

benaadams:
> However, saying that we're not capturing arguments 

You are capturing *all* the arguments; but I assume that's the point?

e.g. `handler` and `statusCodeHandlers` are used in the local function `myFunc` without being passed -however its not capturing any more than the closure (lambda) was previously.

IEvangelist:
> Local functions are allocation-free if no conversion to delegates is happening

Your newly added local functions are simply acting as a replacement for the `RequestDelegate`, as a result these is a conversion from the local function to the delegate. This means that there are allocations, and it is not "allocation-free".

benaadams:
Its only creating them on start up though, so allocations shouldn't matter too much?

jchannon:
No not too much. We also have local functions here https://github.com/jchannon/Botwin/blob/master/src/BotwinModule.cs#L32 but again not sure we've gained by having them.

benaadams:
No captures in the example you just linked; a local function would be better - or moving the lambda to a `readonly static`; as while the lambda will be initalized only once it will still be in the method body which will make the method less likely to be an inline candidate due to its size https://github.com/dotnet/roslyn/issues/13624.

A local function will be better than a `readonly static` lambda as it will be created at compile time, while the `readonly static` will be done at runtime.

Doing it at runtime has the issue that the first method calling it will have an initalization check inserted by the Jit to see whether it has already been created; and everything else calling it will be free of that - though the check is pretty minor in the scheme of things.

benaadams:
I might have been wrong: `handler` :-/

jchannon:
wrong about what?

benaadams:
```csharp
protected void Get(string path, Func<HttpRequest, HttpResponse, RouteData, Task> handler)
{
    Task RequestDelegate(HttpContext httpContext) => handler(httpContext.Request, httpContext.Response, httpContext.GetRouteData());
    this.Get(path, RequestDelegate);
}
```
The lambda is capturing `handler`; so you couldn't even move it to a `readonly static`

jchannon:
Brill. Thanks for the help guys. Will close the PR then as it adds no value.

PR #27, opened 2017-10-10
Added file bind extension methods

JoeStead:
We were getting quite a few extension method classes, moved them all into a folder as well


src/Extensions/BindExtensions.cs
```diff
@@ -0,0 +1,93 @@
...
+                }
+            }
+
+            return postedFiles;
+        }
+        
+        public static async Task BindAndSaveFiles(this HttpRequest request, string saveLocation)
+        {
+            
+            if (request.HasFormContentType)
```
jchannon: This can call BindFiles to get the file list

src/Extensions/BindExtensions.cs
```diff
@@ -0,0 +1,93 @@
...
+                    }
+
+                    postedFiles.Add(file);
+                }
+            }
+
+            return postedFiles;
+        }
+        
+        public static async Task BindAndSaveFiles(this HttpRequest request, string saveLocation)
```
jchannon: I think we need a `BindFile` and `BindAndSaveFile` maybe? In MVC you can inject `IFormFile` or `IEnumerable<IFormFile>` into an Action method. The singular literally calls `.First()` to get the one file

JoeStead:
We should look at getting integration tests running as part of the CI too

src/Extensions/BindExtensions.cs
JoeStead: Added this (by just returning that matches, no point looping through them all).

JoeStead: Then again, this causes duplication a bit 🤔

jchannon:
I think have the loop in a method. BindFiles and BindFile calls it. The
latter just calls .First()

JoeStead:
Yeah, didn't want to do that because the latter will loop through all the files and just take the first one. If we had `yield return` inside an async scope, it'd be fine

jchannon:
ReadFormAsync is the hard part. Unless they are uploading a tonne of files, adding it to a list isn't really going to be that painful.

https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/FormFileModelBinder.cs#L53-L132

JoeStead:
Changed for now, can't say I like it though

jchannon:
It does seem there is a potential DDOS attack there

JoeStead:
Yep, fill the request with thousands of "files" so the server has to loop through them all just to pick the first one.

jchannon:
aye

src/Extensions/BindExtensions.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        public static async Task<IEnumerable<IFormFile>> BindFiles(this HttpRequest request)
+        {
+            var postedFiles = new List<IFormFile>();
+
+            if (request.HasFormContentType)
+            {
+                var form = await request.ReadFormAsync();
+
+                foreach (var file in form.Files)
```
jchannon: Could be ugly but could add a method with optional bool flag `private IEnumerable<IFormFile> GetFiles(bool returnOnFirst = false)` that gets called by BindFile with true set

src/Extensions/BindExtensions.cs
```diff
@@ -61,15 +61,26 @@ public static T Bind<T>(this HttpRequest request)
...
+                    }
                 }
             }
 
             return postedFiles;
         }
 
+        public static async Task<IEnumerable<IFormFile>> BindFiles(this HttpRequest request)
+        {
+            return await request.BindFiles(false);
```
jchannon: set the parameter name? BindFiles(returnOnFirst:false)

JoeStead: It's a private method with one parameter, really?

jchannon:
yup :)

jchannon:
Looks good. What are your thoughts on a later PR to change the namespaces. So the Bind extensions go under Botwin.Binding etc

PR #38, opened 2017-10-15
Fixed route/query parameters binding when using Nullable data-types.

### LINKED ISSUE #37: As<T> throws InvalidCastException on Nullable types
When using `As<T>` and `AsMultiple<T>` extension methods on `IQueryCollection` or `RouteData` with Nullable data-types as `T` (e.g. `int?`), an `InvalidCastException` is thrown.

This is due the framework unable to cast `int` to `int?` (same for other `Nullable<T>`) directly when using `Convert.ChangeType`.

fdipuma:
Fixes #37 

Added internal helper methods for converting types.

Added unit-tests for nullable binding on query string.

### DIFF
--- a/src/Request/ConvertExtensions.cs
+++ b/src/Request/ConvertExtensions.cs
@@ -1,0 +1,42 @@
+﻿namespace Botwin.Request
+{
+    using System;
+    using System.Collections.Generic;
+
+    internal static class ConvertExtensions
+    {
+        public static T ConvertTo<T>(this object value)
+        {
+            var type = typeof(T);
+            var underlyingType = Nullable.GetUnderlyingType(type);
+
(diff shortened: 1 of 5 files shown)


src/Request/ConvertExtensions.cs
```diff
@@ -0,0 +1,42 @@
...
+            {
+                if (value == null) return default(T);
+
+                type = underlyingType;
+            }
+
+            return (T)Convert.ChangeType(value, type);
+        }
+
+        public static IEnumerable<T> ConvertMultipleTo<T>(this IEnumerable<object> values)
```
jchannon: I think this can be `IEnumerable<string>` for now

jchannon:
Great stuff, thanks! 👍

PR #42, opened 2017-10-19
Passing cancellation token into IResponseNegotiator

jasonmitchell:
Methods in `ResponseExtensions` had an optional parameter but this wasn't used.  I've added a cancellation token to `IResponseNegotiator.Handle` and passed this into the call to `WriteAsync`.


jchannon:
Thanks!! 👍

Issue #44, opened 2017-10-20
Add XML docs to public methods


jchannon:
- [x] BindExtensions - done via #46 
- [x] ValidationExtensions - done via #47 
- [x] QueryStringExtensions - done via #47
- [x] RouteDataExtensions - done via #49 
- [x] StreamExtensions - done via #48
- [x] ResponseExtensions - done via #49 
- [x] BotwinExtensions - done via #50 
- [x] BotwinModule - done via #55 
- [x] BotwinOptions - done via #56 
- [x] IValidatorLocator - done via #57 
- [x] IStatusCodeHandler - done via #58 
- [x] IResponseNegotiator - done via #49

PR #49, opened 2017-10-31
Xml docs for IResponseNegotiator, RouteDataExtensions, ResponseExtensions

pdwetz:
A few more xml doc additions for #44


src/Response/ResponseExtensions.cs
```diff
@@ -10,6 +10,13 @@ namespace Botwin.Response
 
     public static class ResponseExtensions
     {
+        /// <summary>
+        /// Handles current response, utilizing an accepted media type if possible and defaulting to "application/json" if none found.
```
jchannon: `Executes content negotiation on current <see cref="HttpResponse"/>, utilizing...`

src/Response/ResponseExtensions.cs
```diff
@@ -21,6 +28,13 @@ public static class ResponseExtensions
             await negotiator.Handle(response.HttpContext.Request, response, obj, cancellationToken);
         }
 
+        /// <summary>
+        /// Handles current response as Json data
```
jchannon: Returns a Json response

jchannon:
Thanks so much! Great work 👍

PR #52, opened 2017-11-02
add 405 fix handler

### LINKED ISSUE #51: Check Botwin returns 405 properly unlike MVC
See here https://github.com/aspnet/Mvc/issues/388

jchannon:
Fixes #51

### DIFF
--- a/src/BotwinExtensions.cs
+++ b/src/BotwinExtensions.cs
@@ -19,14 +19,16 @@ namespace Botwin
 
         public static IApplicationBuilder UseBotwin(this IApplicationBuilder builder, BotwinOptions options)
         {
+            //Invoke so ctors are called that adds routes to IRouter
+            var srvs = builder.ApplicationServices.GetServices<BotwinModule>();
+
+            Add405Handler(builder, srvs);
+
             ApplyGlobalBeforeHook(builder, options);
 
             ApplyGlobalAfterHook(builder, options);
 
             var routeBuilder = new RouteBuilder(builder);
-
-            //Invoke so ctors are called that adds routes to IRouter
-            var srvs = builder.ApplicationServices.GetServices<BotwinModule>();
 
             //Cache status code handlers
             var statusCodeHandlers = builder.ApplicationServices.GetServices<IStatusCodeHandler>();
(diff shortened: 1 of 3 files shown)


src/BotwinExtensions.cs
```diff
@@ -40,15 +42,38 @@ public static IApplicationBuilder UseBotwin(this IApplicationBuilder builder, Bo
...
                 }
             }
 
             return builder.UseRouter(routeBuilder.Build());
         }
 
-        
+        private static void Add405Handler(IApplicationBuilder builder, IEnumerable<BotwinModule> srvs)
+        {
+            var systemRoutes = new List<(string verb, string route)>();
```
JoeStead: `Dictionary<path, IEnumerable<verb>` would probably read nicer in this method, and might be quicker than all of the linq too

jchannon: to load the dictionary you'd have to do ` systemRoutes.Add(route.path, module.Routes.Where(x=>x.path == route.path).Select(x=>x.verb));`

i think this whole piece of code could be improved though

JoeStead: Hmm, you're not wrong. There's gotta be a nicer way of getting that code in there

fdipuma: Wouldn't something like this be more efficient in the long run?

```
var systemRoutes = srvs.SelectMany(m => m.Routes)
    .GroupBy(r => "/" + r.path)
    .ToDictionary(
        g => g.Key, 
        g => new HashSet<string>(
            g.Select(r => r.verb).Distinct()
        )
    ); // IDictionary<string, HashSet<string>>
```

We than could just do:
```
if (systemRoutes.ContainsKey(strippedPath) &&
    !systemRoutes[strippedPath].Contains(context.Request.Method)) // almost O(1) access time
{
    context.Response.StatusCode = 405;
    return;
}
```

(the LINQ part may be simplified into a `foreach` if it appears too cryptical)

jchannon:
I did play with something similar after this PR.

My actual concern is the string comparison of the paths for example a route
might be defined as /home/ but the route that comes in might be /home or
vice versa. Only way I can think is to strip both the registered path and
the incoming path of the end slash but that seems inefficient to me but
might be maybe only option we have

jchannon:
Pushed a change which should make the checks more precise but still not happy about the stripping per path on each request approach

jchannon:
@davidfowl any better perf suggestions?

jchannon:
Moved to #65

PR #62, opened 2017-11-14
Register and resolve modules from request scope

### LINKED ISSUE #61: Question about services lifecycles
I´m trying to use the Botwin library with EF Core. I registered the DbContext using services.AddDbContext (By default sets ServiceLifetime.Scoped) but I noticed that my application uses the same DbContext in each request. Then I noticed that the modules are created once, regardless of whether they are registered at IoC container as transient. So each dependency injected don't respect its configured Lifetime. Is this a desing decision?

Thanks in advance.

jchannon:
Fixes #61

### DIFF
--- a/src/BotwinExtensions.cs
+++ b/src/BotwinExtensions.cs
@@ -47,7 +47,7 @@ namespace Botwin
 
                     foreach (var route in module.Routes.Keys)
                     {
-                        routeBuilder.MapVerb(route.verb, route.path, ctx =>
+                        routeBuilder.MapVerb(route.verb, route.path, async ctx =>
                         {
                             var statusCodeHandlers = ctx.RequestServices.GetServices<IStatusCodeHandler>();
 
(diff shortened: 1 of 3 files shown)


jchannon:
I think this is going to be as good as it gets for a first pass. As @khellang said there will be a lot of allocations per request with this though. I can't see how you can cache anything to improve perf here. @davidfowl, do you have any thoughts before this gets merged?

src/BotwinExtensions.cs
```diff
@@ -164,13 +172,14 @@ public static void AddBotwin(this IServiceCollection services, params Assembly[]
             var modules = assemblies.SelectMany(x => x.GetTypes().Where(t => typeof(BotwinModule).IsAssignableFrom(t) && t != typeof(BotwinModule)));
             foreach (var module in modules)
             {
-                services.AddTransient(typeof(BotwinModule), module);
+                services.AddScoped(module);
```
davidfowl: Why not leave it as transient?

jchannon:
Because the modules are not used once the route is registered with
Routebuilder. If you create a new guid in the ctor and return it in the
route it is always the same value. The ctor is only hit once at startup not
once per request so all dependencies currently become a singleton

davidfowl:
Yea but you aren't fixing anything by making a scope. You're still creating the modules at app startup time. You don't need to make anything scoped here, they're probably only resolved once per request right? 

You just need to use the service provider on the HttpContext to resolve the module again inside of the route callback. That should be enough to solve your problems.

jchannon:
@davidfowl Hmm OK, will take a look thanks. @khellang any thoughts?

jchannon:
Removing the scope does still work with the module being resolved inside the wrapping handler

davidfowl:
Hmm actually it'll break if you inject a DbContext (or any other scoped dependency). Go ahead and keep the scope for now. It's hacky but it solves this problem.

The biggest downside is the fact that you have to instantiate the module, run the constructor which to get the list of routes which are only used once at startup. You always create a single instance of a module that gets thrown away.

jchannon:
@davidfowl is there any way to improve the perf doing that resolving, creating the actual handler each request?

davidfowl:
I forgot to add that work also happens per request but by then nobody cares about the routes.

jchannon:
Well the routes are looked up from the singleton module by the request scoped module to find the correct handler

khellang:
> Hmm actually it'll break if you inject a DbContext (or any other scoped dependency).

Yeah, the scope has to be there in order to not leak scoped dependencies resolved from the app container...

src/BotwinExtensions.cs
jchannon: @khellang why is the module added twice?

poke: @jchannon You need to add the module once as `BotwinModule`, so you can resolve all modules by requesting `IEnumerable<BotwinModule>` from DI. And then, you need to register each module as itself, so you can get the exact module later when you retrieve the module in the path handler.

jchannon: Seems odd 😄

poke: Maybe, but basically you are just assigning each module two “labels” you can refer it to. Once as the module itself, and once with it being *a* module.

jchannon:
Thanks @poke @khellang @davidfowl 👍 ❤️ 👍

PR #63, opened 2017-11-15
When registering modules, check if the class is not abstract. This al…

jonathanschein:
…lows to include abstract base classes to generalize code

I added this small change because if I had some module abstract super class, the AddBotwin method was trying to add that to the service collection.


jchannon:
Thanks! 👍

PR #64, opened 2017-11-16
Continuing module resolution from request scope (#62)

poke:
This continues the work started by @khellang in #62.

In order to allow for better route lookup, this changes the route list into a dictionary in the module (as already mentioned before in chat). That way, you can now look up a route handler by verb and path instead of having to iterate over the list for a match. In addition, this also has the side effect, that you cannot register multiple handlers for identical routes anymore (maybe this needs some better error handling in the module methods though).

The next step was to avoid creating all those handlers when a route is being activated. The solution is actually pretty simple if you think about what was going on before: All that was happening was that the originally registered route handler was wrapped twice, once with the before and after handler from the module, and then again with that final thing. But we don’t actually need to wrap anything here; instead, we can just combine all those handlers into a single one that does everything—so we basically inline all those wrapping. The result is a single handler that is being constructed *at startup time*, so we are not allocating any handler when the route is later being executed. We just call the correct things in the correct order.

Finally, I refactored that a bit since I was hitting a naming issue myself where I used both the `requestScopedModule` and the original `module` within the handler. To avoid this from happening, I extracted this out in its own method so that you have a clear interface (the method signature) about what will be captured within the created lambda expression.

### DIFF
--- a/src/BotwinExtensions.cs
+++ b/src/BotwinExtensions.cs
@@ -47,7 +47,7 @@ namespace Botwin
 
                     foreach (var route in module.Routes.Keys)
                     {
-                        routeBuilder.MapVerb(route.verb, route.path, ctx =>
+                        routeBuilder.MapVerb(route.verb, route.path, async ctx =>
                         {
                             var statusCodeHandlers = ctx.RequestServices.GetServices<IStatusCodeHandler>();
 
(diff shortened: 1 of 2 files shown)


src/BotwinExtensions.cs
```diff
@@ -44,41 +45,52 @@ public static IApplicationBuilder UseBotwin(this IApplicationBuilder builder, Bo
...
                     //Cannot read the default stream once WriteAsync has been called on it
                     ctx.Response.Body = new MemoryStream();
                 }
 
-                await handler(ctx);
+                // run the module handlers
+                bool shouldContinue = true;
+                if (module.Before != null)
+                {
+                    shouldContinue = await module.Before(ctx);
```
jchannon: I think this should return at this point if `shouldContinue` is false?

src/BotwinExtensions.cs
```diff
@@ -44,41 +45,52 @@ public static IApplicationBuilder UseBotwin(this IApplicationBuilder builder, Bo
...
         }
 
-        private static RequestDelegate CreateFinalHandler(RequestDelegate handler, IEnumerable<IStatusCodeHandler> statusCodeHandlers)
+        private static RequestDelegate CreateRouteHandler((string verb, string path) route, Type moduleType)
         {
-            RequestDelegate finalHandler = async (ctx) =>
+            return async (HttpContext ctx) =>
             {
+                var module = ctx.RequestServices.GetRequiredService(moduleType) as BotwinModule;
+                if (!module.Routes.TryGetValue((route.verb, route.path), out var routeHandler))
```
khellang: I guess this makes both verb (method) and path case sensitive? Is that OK?

src/BotwinExtensions.cs
poke: I originally had this to return, however I realized that in the original implementation, the `return` only affected the “before and after handler”. So returning here would be wrong as it would terminate the full handler, including what happened inside the final handler and e.g. the status code handlers.

So returning `false` from the before handler will prevent the module’s route handler from running (and the after handler) but not terminate the response immediately to allow the other components from finishing their job.

khellang:
Just an FYI; [the RFC](https://tools.ietf.org/html/rfc7231#section-4.1) says methods are case sensitive

> The method token is case-sensitive because it might be
   used as a gateway to object-based systems with case-sensitive method
   names.

But I don't think many implementations enforce this, including ASP.NET Core.

src/BotwinExtensions.cs
```diff
@@ -44,41 +45,52 @@ public static IApplicationBuilder UseBotwin(this IApplicationBuilder builder, Bo
...
+                    throw new InvalidOperationException($"Route {route.verb} '{route.path}' was no longer found");
```
khellang: You'd have to do some pretty evil things for this to happen 😂

jchannon:
Sales Pitch, Botwin is a RFC7231 enforcer 😄

src/BotwinExtensions.cs
poke: Yeah, you’re right. If that’s not desired, we should pass a different `IEqualityComparer` to the dictionary constructor in `BotwinModule`. That way, it would still keep the lookup benefits.

For string keys, we could use the built-in `StringComparer.InvariantCultureIgnoreCase`. For the two-tuple key, I guess we will have to run our own comparer, e.g. as shown [here](https://stackoverflow.com/questions/16428492/case-insensitive-dictionary-with-tuple-key). So it would probably look like this:

```
class TwoTupleStringComparer : IEqualityComparer<ValueTuple<string, string>>
{
    public bool Equals(ValueTuple<string, string> x, ValueTuple<string, string> y)
        => StringComparer.InvariantCultureIgnoreCase.Equals(x.Item1, y.Item1) && StringComparer.InvariantCultureIgnoreCase.Equals(x.Item2, y.Item2);

    public int GetHashCode(ValueTuple<string, string> obj)
        => (StringComparer.InvariantCultureIgnoreCase.GetHashCode(obj.Item1), StringComparer.InvariantCultureIgnoreCase.GetHashCode(obj.Item2)).GetHashCode();
}
```

src/BotwinExtensions.cs
poke: Yeah, I know xD

src/BotwinExtensions.cs
khellang: I'm not sure if using case-insensitive comparison for the path would be correct. What does Botwin do already? I guess the old check actually _was_ case-senstitive (`==`).

PR #65, opened 2017-11-16
Add 405 handler

### LINKED ISSUE #51: Check Botwin returns 405 properly unlike MVC

jchannon:
Fixes #51

### DIFF
--- a/src/BotwinExtensions.cs
+++ b/src/BotwinExtensions.cs
@@ -1,6 +1,7 @@
 namespace Botwin
 {
     using System;
+    using System.Collections.Generic;
     using System.IO;
     using System.Linq;
     using System.Reflection;
@@ -39,8 +40,12 @@ namespace Botwin
             //Create a "startup scope" to resolve modules from
             using (var scope = builder.ApplicationServices.CreateScope())
             {
+                var modules = scope.ServiceProvider.GetServices<BotwinModule>();
+
+                Apply405Handler(builder, modules);
+
                 //Get all instances of BotwinModule to fetch and register declared routes
-                foreach (var module in scope.ServiceProvider.GetServices<BotwinModule>())
+                foreach (var module in modules)
                 {
                     var moduleType = module.GetType();
 
@@ -52,6 +57,34 @@ namespace Botwin
             }
 
             return builder.UseRouter(routeBuilder.Build());
+        }
+
+        private static void Apply405Handler(IApplicationBuilder builder, IEnumerable<BotwinModule> modules)
+        {
+            var systemRoutes = new List<(string verb, string route)>();
+            foreach (var module in modules)
+            {
+                foreach (var route in module.Routes.Keys)
+                {
+                    var strippedPath = route.path.EndsWith("/") ? route.path.Substring(0, route.path.Length - 1) : route.path;
+                    systemRoutes.Add((route.verb, "/" + strippedPath));
+                }
+            }
+
+            builder.Use(async (context, next) =>
+            {
+                var strippedPath = context.Request.Path.Value.EndsWith("/") && context.Request.Path.Value.Length > 1
+                    ? context.Request.Path.Value.Substring(0, context.Request.Path.Value.Length - 1)
+                    : context.Request.Path.Value;
+
+                var verbsForPath = systemRoutes.Where(x => x.route == strippedPath).Select(y => y.verb);
+                if (verbsForPath.All(x => x != context.Request.Method))
+                {
+                    context.Response.StatusCode = 405;
+                    return;
+                }
+                await next();
+            });
         }
 
         private static RequestDelegate CreateRouteHandler((string verb, string path) route, Type moduleType)
--- a/src/BotwinExtensions.cs
+++ b/src/BotwinExtensions.cs
@@ -5,6 +5,7 @@ namespace Botwin
     using System.IO;
     using System.Linq;
     using System.Reflection;
+    using System.Threading.Tasks;
     using FluentValidation;
     using Microsoft.AspNetCore.Builder;
     using Microsoft.AspNetCore.Http;
@@ -36,13 +37,12 @@ namespace Botwin
             ApplyGlobalAfterHook(builder, options);
 
             var routeBuilder = new RouteBuilder(builder);
+            var systemRoutes = new List<(string verb, string route)>();
 
             //Create a "startup scope" to resolve modules from
             using (var scope = builder.ApplicationServices.CreateScope())
             {
                 var modules = scope.ServiceProvider.GetServices<BotwinModule>();
-
-                Apply405Handler(builder, modules);
 
                 //Get all instances of BotwinModule to fetch and register declared routes
                 foreach (var module in modules)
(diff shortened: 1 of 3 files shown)


jchannon:
Pinging @poke to cast his wise eyes over

src/BotwinExtensions.cs
```diff
@@ -54,6 +59,34 @@ public static IApplicationBuilder UseBotwin(this IApplicationBuilder builder, Bo
             return builder.UseRouter(routeBuilder.Build());
         }
 
+        private static void Apply405Handler(IApplicationBuilder builder, IEnumerable<BotwinModule> modules)
+        {
+            var systemRoutes = new List<(string verb, string route)>();
+            foreach (var module in modules)
+            {
+                foreach (var route in module.Routes.Keys)
+                {
```
poke: We’re already iterating the module routes in line 48, so we could just combine collecting all “system routes” with setting up the normal route handler here, to avoid iterating everything twice.

src/BotwinExtensions.cs
```diff
@@ -54,6 +59,34 @@ public static IApplicationBuilder UseBotwin(this IApplicationBuilder builder, Bo
...
+                    var strippedPath = route.path.EndsWith("/") ? route.path.Substring(0, route.path.Length - 1) : route.path;
+                    systemRoutes.Add((route.verb, "/" + strippedPath));
+                }
+            }
+
+            builder.Use(async (context, next) =>
```
poke: If I understand it correctly, then this handler is supposed to “pick up” those routes which paths are registered but are using the incorrect verb? In that case, wouldn’t it be better to run this middleware *after* the normal mapped routes, so it only runs when handling those is over (similarly how in standard ASP.NET, MVC runs *after* StaticFiles *if* StaticFiles could not respond to a request).

src/BotwinExtensions.cs
```diff
@@ -54,6 +59,34 @@ public static IApplicationBuilder UseBotwin(this IApplicationBuilder builder, Bo
...
+            {
+                var strippedPath = context.Request.Path.Value.EndsWith("/") && context.Request.Path.Value.Length > 1
+                    ? context.Request.Path.Value.Substring(0, context.Request.Path.Value.Length - 1)
+                    : context.Request.Path.Value;
+
+                var verbsForPath = systemRoutes.Where(x => x.route == strippedPath).Select(y => y.verb);
```
poke: If you’re looking up routes by paths, then you shouldn’t use a list. Consider an `ILookup<string, string>` instead that maps from paths to the registered verbs.

src/BotwinExtensions.cs
```diff
@@ -54,6 +59,34 @@ public static IApplicationBuilder UseBotwin(this IApplicationBuilder builder, Bo
...
+                if (verbsForPath.All(x => x != context.Request.Method))
```
poke: Per above: If this would run *after* the normal handlers as a fallback, then at this point you would be already guaranteed that this is running on an unsupported method. So you wouldn’t actually need to lookup (or even store) the registered methods but just the paths.

src/BotwinExtensions.cs
jchannon: If it runs after, the routing middleware will set 404 as the status code, the issue is, how do you know its a 404 from an unfound route vs a 404 wrongly assigned for invalid verb on path?

src/BotwinExtensions.cs
jchannon: I initially tried to do that but i need to loop to get all routes before applying the 405 handler which is global whereas the route handler is per route

poke: I was thinking about something like this:

```
foreach (var route in …)
{
    routeBuilder.MapVerb(…);
    knownRoutes.Add(route);
}

builder.UseRouter(routeBuilder.Build());
builder.Use(GetMethodNotAllowedHandler(knownRoutes));
```

src/BotwinExtensions.cs
poke: Unless I’m completely wrong (routing is one of the topics I didn’t really do anything with yet) but from [how it looks](https://github.com/aspnet/Routing/blob/rel/2.0.0/src/Microsoft.AspNetCore.Routing/RouterMiddleware.cs#L29), the next middleware is only invoked if there is no matching route. So if the middleware is *after* the router, it will only run for unrecognized routes. So you would only have to check whether *any* route with such a path exists in order to determine whether it’s a real not found or just an incorrect verb. So you could do this with a simple hash set on the paths.

src/BotwinExtensions.cs
```diff
@@ -35,23 +37,58 @@ public static IApplicationBuilder UseBotwin(this IApplicationBuilder builder, Bo
...
-                foreach (var module in scope.ServiceProvider.GetServices<BotwinModule>())
+                foreach (var module in modules)
                 {
                     var moduleType = module.GetType();
 
                     foreach (var route in module.Routes.Keys)
                     {
                         routeBuilder.MapVerb(route.verb, route.path, CreateRouteHandler(route, moduleType));
+
+                        var strippedPath = route.path.EndsWith("/") ? route.path.Substring(0, route.path.Length - 1) : route.path;
```
poke: Not sure what David would say about this (allocations!?!?!), but maybe just `route.path.TrimEnd('/')`?

src/BotwinExtensions.cs
```diff
@@ -35,23 +37,58 @@ public static IApplicationBuilder UseBotwin(this IApplicationBuilder builder, Bo
...
+                        systemRoutes.Add((route.verb, "/" + strippedPath));
```
poke: It might be cheaper to leave off the leading slash here, and trim the leading slash from the current request later instead.

src/BotwinExtensions.cs
```diff
@@ -35,23 +37,58 @@ public static IApplicationBuilder UseBotwin(this IApplicationBuilder builder, Bo
...
+        }
+
+        /// <summary>
+        /// This method is only called if it's a valid 404 or 405
+        /// </summary>
+        /// <param name="context"></param>
+        /// <param name="next"></param>
+        /// <param name="systemRoutes"></param>
+        /// <returns></returns>
+        private static async Task GetMethodNotAllowedHandler(HttpContext context, Func<Task> next, IEnumerable<(string verb, string route)> systemRoutes)
```
poke: Since this does not return a handler function (which is what I was originally suggesting) but *is* the handler, the `Get` feels out of place here. `HandleMethodNotAllowed`? Or maybe you should just move this out to a custom middleware class.

src/BotwinExtensions.cs
```diff
@@ -35,23 +37,58 @@ public static IApplicationBuilder UseBotwin(this IApplicationBuilder builder, Bo
...
+        {
+            //Call the final pipeline which gets ASP.Net Core status code, usually a 404 in this case
+            await next();
```
poke: Why are you continuing the pipeline here at the beginning? Shouldn’t this handler already be kind of the final one in the pipeline? Do you need to know that it produced a 404? Wouldn’t it be safe to assume that if this handler runs after the routing middleware, then a route was not found? (again: I have no actual idea about the routing middleware)

src/BotwinExtensions.cs
```diff
@@ -35,23 +37,58 @@ public static IApplicationBuilder UseBotwin(this IApplicationBuilder builder, Bo
...
 
-            return builder.UseRouter(routeBuilder.Build());
+            var strippedPath = context.Request.Path.Value.EndsWith("/") && context.Request.Path.Value.Length > 1
+                ? context.Request.Path.Value.Substring(0, context.Request.Path.Value.Length - 1)
+                : context.Request.Path.Value;
+
+            //ASP.Net Core will set a 405 response to 404. Let's check if it's a valid 404 first otherwise if we know about the route it's most likely a 405
+            if (context.Response.StatusCode == 404 && systemRoutes.Any(x => x.route == strippedPath))
```
poke: Use HashSet or Dictionary for constant time lookup. Also, I don’t know if the router middleware is case insensitive itself but if it is, then pure string equality will not work (you will have to use the RouteComparer we came up with earlier).

src/BotwinExtensions.cs
jchannon: Yeah tried your approach but no joy

src/BotwinExtensions.cs
jchannon: Oddly before that is called the response status code is 200, if you call it next it's then 404 from the router. Not sure how because as you say I'd expect it to be already at the end of the pipeline

poke: I don’t think you actually need to check the status code here though. This should not run if the router succeeded in the pipeline step before.

jchannon: Interestingly if I remove the `next()` call it will always be 200 on a route not defined in Botwin. If I take the GetMethodNotAllowedHandler out completely I get a 404 so I assume there must be one last pipeline called somewhere

jchannon:
Fixed in #66

PR #66, opened 2017-11-19
Implement 405 Method Not Allowed handling (#51)

poke:
So I had an epiphany this night and just needed to get this out… xD

This implements the 405 Method Not Allowed handling, fixing #51. This should be a very efficient solution that has no actual drawbacks *and* it also supports parameterized routes which was the blocking part in the previous attempt.

Only limitation is that this will not work properly when multiple modules register the same *paths* but with different verbs. But I think that’s not really a desired use case anyway?

### DIFF
--- a/src/BotwinExtensions.cs
+++ b/src/BotwinExtensions.cs
@@ -1,9 +1,11 @@
 namespace Botwin
 {
     using System;
+    using System.Collections.Generic;
     using System.IO;
     using System.Linq;
     using System.Reflection;
+    using System.Threading.Tasks;
     using FluentValidation;
     using Microsoft.AspNetCore.Builder;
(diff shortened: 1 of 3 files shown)


jchannon:
LEGEND

jchannon:
Thanks!

poke:
Hah, you’re welcome :)

Btw. I also thought about making this optional behavior by putting some attribute onto the module class. That way you could make this opt-in and add it without causing a breaking change. Just in case that’s something you would want :)

jchannon:
No that's fine, will add it as a breaking change. Only thing would be to
try and somehow get the behaviour where routes are defined in different
modules although I think user behaviour is rare for that so will deal with
it if it becomes an issue based on feedback

PR #72, opened 2017-12-01
Fixed #71 and added functional non-ioc sample

### LINKED ISSUE #71: Why aren't exceptions handled by the Exception middleware
In the samples folder if you hit /error it throws an exception.  The `UseExceptionHandler ` middleware should capture that and redirect.  This used to work but no longer does, we need to investigate why.

jchannon:


samples/FunctionalDependencyModule.cs
```diff
@@ -0,0 +1,54 @@
...
+    using System.Linq;
+    using Botwin.Response;
+
+    public class FunctionalDependencyModule : BotwinModule
+    {
+        public FunctionalDependencyModule()
+        {
+            this.Get("/functional", async (req, res, routeData) =>
+            {
+                var handler = Composition.ComposeFunctionalHandler();
```
jchannon: How do we replace this in an acceptance in mem http pipeline test so we remove live sql connections

PR #78, opened 2017-12-12
Extract Validate method

### LINKED ISSUE #77: Separate BindAndValidate so Validate can be called on its own
Currently, the `BindAndValidate` method does exactly what it says, but there is no way to validate the payload without binding it. 

This should just be a case of introducing a public `Validate` method, can calling that method from  `BindAndValidate`

jussimattila:
Closes #77 

Moved model validation code from within `BindAndValidate` to a separate `Validate` extension method. It is a bit weird having the new method extend `HttpRequest`, which is only needed for service locator, but I left the code as it was.

I modified docs slightly. I didn't add tests since the code is still covered by existing tests for `BindAndValidate`.


jchannon:
Thanks!! 👍

PR #83, opened 2018-02-16
Fixed published assembly issue

### LINKED ISSUE #82: Sample release version doesn't work
there is no problem with Debug mode.
but in release and publish mode, routing doesn't seems work.

jchannon:
Fixes #82

### DIFF
--- a/src/BotwinExtensions.cs
+++ b/src/BotwinExtensions.cs
@@ -1,7 +1,6 @@
 namespace Botwin
 {
     using System;
-    using System.Collections.Generic;
     using System.IO;
     using System.Linq;
     using System.Reflection;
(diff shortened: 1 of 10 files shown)


src/BotwinExtensions.cs
```diff
@@ -148,9 +147,11 @@ private static void ApplyGlobalBeforeHook(IApplicationBuilder builder, BotwinOpt
         /// </summary>
         /// <param name="services">The <see cref="IServiceCollection"/> to add Botwin to.</param>
         /// <param name="assemblies">Optional array of <see cref="Assembly"/> to add to the services collection. If assemblies are not provided, Assembly.GetEntryAssembly is called.</param>
-        public static void AddBotwin(this IServiceCollection services, params Assembly[] assemblies)
+        public static void AddBotwin(this IServiceCollection services)
```
JoeStead: We should probably provide an overload so you can specify the assembly your modules belong in explicitly.

JoeStead: Actually, I don't think you do need this

PR #87, opened 2018-03-08
Add trace logging

### LINKED ISSUE #21: Use the ILogger stuff to add debug level tracing
As we're using the `IServiceProvider` provided by AspNetCore, we should hook up some of the other features, including enabling logging.

It's a pain to try and diagnose issues with things without any logging, we should implement trace/debug calls around the code base that can be configured with the ILoggerFactory

jchannon:
Fixes #21 

When running with log level trace the following output is from starting the app and processing one request

```
trce: Botwin.BotwinExtensions[0]
      Found Botwin.Samples.ActorValidator
trce: Botwin.BotwinExtensions[0]
      Found Botwin.Samples.DirectorValidator
trce: Botwin.BotwinExtensions[0]
      Found ValidatorOnlyProject.CastMemberValidator
trce: Botwin.BotwinExtensions[0]
      Found Botwin.Samples.ActorsModule
trce: Botwin.BotwinExtensions[0]
      Found Botwin.Samples.ErrorModule
trce: Botwin.BotwinExtensions[0]
      Found Botwin.Samples.FunctionalProgrammingModule
trce: Botwin.BotwinExtensions[0]
      Found Botwin.Samples.HomeModule
trce: Botwin.BotwinExtensions[0]
      Found Botwin.Samples.HooksModule
trce: Botwin.BotwinExtensions[0]
      Found BotwinSample.Features.Directors.CastMemberModule
trce: Botwin.BotwinExtensions[0]
      Found Botwin.Samples.ConflictStatusCodeHandler
trce: Botwin.BotwinExtensions[0]
      Found Botwin.DefaultJsonResponseNegotiator
trce: Botwin.BotwinExtensions[0]
      Adding Botwin to application
trce: Botwin.BotwinExtensions[0]
      Applying global before hook
trce: Botwin.BotwinExtensions[0]
      Applying global after hook
trce: Botwin.BotwinExtensions[0]
      Found and adding ActorsModule with GET /actors to routing
trce: Botwin.BotwinExtensions[0]
      Found and adding ActorsModule with HEAD /actors to routing
trce: Botwin.BotwinExtensions[0]
      Found and adding ActorsModule with GET /actors/{id:int} to routing
trce: Botwin.BotwinExtensions[0]
      Found and adding ActorsModule with HEAD /actors/{id:int} to routing
trce: Botwin.BotwinExtensions[0]
      Found and adding ActorsModule with PUT /actors/{id:int} to routing
trce: Botwin.BotwinExtensions[0]
      Found and adding ActorsModule with POST /actors to routing
trce: Botwin.BotwinExtensions[0]
      Found and adding ErrorModule with GET /error to routing
trce: Botwin.BotwinExtensions[0]
      Found and adding ErrorModule with HEAD /error to routing
trce: Botwin.BotwinExtensions[0]
      Found and adding ErrorModule with GET /errorhandler to routing
trce: Botwin.BotwinExtensions[0]
      Found and adding ErrorModule with HEAD /errorhandler to routing
trce: Botwin.BotwinExtensions[0]
      Found and adding FunctionalProgrammingModule with GET /functional/directors to routing
trce: Botwin.BotwinExtensions[0]
      Found and adding FunctionalProgrammingModule with HEAD /functional/directors to routing
trce: Botwin.BotwinExtensions[0]
      Found and adding FunctionalProgrammingModule with GET /functional/directors/{id:int} to routing
trce: Botwin.BotwinExtensions[0]
      Found and adding FunctionalProgrammingModule with HEAD /functional/directors/{id:int} to routing
trce: Botwin.BotwinExtensions[0]
      Found and adding FunctionalProgrammingModule with POST /functional/directors to routing
trce: Botwin.BotwinExtensions[0]
      Found and adding FunctionalProgrammingModule with PUT /functional/directors/{id:int} to routing
trce: Botwin.BotwinExtensions[0]
      Found and adding FunctionalProgrammingModule with DELETE /functional/directors/{id:int} to routing
trce: Botwin.BotwinExtensions[0]
      Found and adding HomeModule with GET / to routing
trce: Botwin.BotwinExtensions[0]
      Found and adding HomeModule with HEAD / to routing
trce: Botwin.BotwinExtensions[0]
      Found and adding HooksModule with GET /hooks to routing
trce: Botwin.BotwinExtensions[0]
      Found and adding HooksModule with HEAD /hooks to routing
trce: Botwin.BotwinExtensions[0]
      Found and adding CastMemberModule with POST /castmembers to routing
dbug: Microsoft.AspNetCore.Hosting.Internal.WebHost[3]
      Hosting starting
dbug: Microsoft.AspNetCore.Server.Kestrel[0]
      No listening endpoints were configured. Binding to http://localhost:5000 by default.
dbug: Microsoft.AspNetCore.Hosting.Internal.WebHost[4]
      Hosting started
dbug: Microsoft.AspNetCore.Hosting.Internal.WebHost[0]
      Loaded hosting startup assembly BotwinSample
Hosting environment: Production
Content root path: /Users/jonathan/Projects/Botwin/samples/BotwinSample
Now listening on: http://localhost:5000
Application started. Press Ctrl+C to shut down.
dbug: Microsoft.AspNetCore.Server.Kestrel[1]
      Connection id "0HLC4V9Q2F2SV" started.
info: Microsoft.AspNetCore.Hosting.Internal.WebHost[1]
      Request starting HTTP/1.1 GET http://localhost:5000/  
trce: Botwin.BotwinExtensions[0]
      Executing global before hook
trce: Botwin.BotwinExtensions[0]
      Executing next handler after global before hook
dbug: Microsoft.AspNetCore.Routing.RouteBase[1]
      Request successfully matched the route with name '(null)' and template ''.
trce: Botwin.BotwinExtensions[0]
      Executing module route handler for GET /
trce: Botwin.BotwinExtensions[0]
      Executing module after hook
Catch you later!
trce: Botwin.BotwinExtensions[0]
      Executing IStatusCodeHandler ConflictStatusCodeHandler
trce: Botwin.BotwinExtensions[0]
      Executing global after hook
We hit a route!
dbug: Microsoft.AspNetCore.Server.Kestrel[9]
      Connection id "0HLC4V9Q2F2SV" completed keep alive response.
info: Microsoft.AspNetCore.Hosting.Internal.WebHost[2]
      Request finished in 178.995ms 409 
dbug: Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv[6]
      Connection id "0HLC4V9Q2F2SV" received FIN.
dbug: Microsoft.AspNetCore.Server.Kestrel[10]
      Connection id "0HLC4V9Q2F2SV" disconnecting.
dbug: Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv[7]
      Connection id "0HLC4V9Q2F2SV" sending FIN.
dbug: Microsoft.AspNetCore.Server.Kestrel[2]
      Connection id "0HLC4V9Q2F2SV" stopped.


```


JoeStead:
About time :)

jchannon:
So as @davidfowl pointed out there is no nice way to log Add* extension methods this PR adds ILogger/ILoggerFactory to the Add extension.  It also does that to the UseBotwin extension.  The sample project changes shows how you might implement the tracing by injecting ILogger into the Startup class

davidfowl:
Why don't you do what MVC does and add the IRouter to the DI container? 

https://github.com/aspnet/Mvc/blob/760c8f38678118734399c58c2dac981ea6e47046/src/Microsoft.AspNetCore.Mvc.Core/Builder/MvcApplicationBuilderExtensions.cs#L91-L100

You can avoid all of the ugly explicit passing of the ILoggerFactory to the UseBotwin method.

jchannon:
I do? https://github.com/jchannon/Botwin/blob/master/src/BotwinExtensions.cs#L56

davidfowl:
Then why are you passing an ILogger here https://github.com/jchannon/Botwin/pull/87/files#diff-82112b99eae34877f111696bbe53aa64R25?

davidfowl:
IMO this is how botwin wire up should look:

```C#
services.AddBotwin(); // This is the default
```

```C#
services.AddBotwin(options => 
{
      // options is typed as BotwinOptions. Under the covers it calls services.Configure<BotwinOptions>,
      // which you can read in the middleware by taking in IOptions<BotwinOptions>
}); 
```

Get rid of the logging during AddBotwin and defer that logging to another time when you can create the logger yourself and not have the user pass it in.

jchannon:
I was going to do that but didn't realise about the options approach. I was
thinking the only way would be for the user to inject
ILogger/ILoggerFactory into their Startup and then pass it to Botwin.

Will try your approach

Thanks :)

jchannon:
Just to confirm are you talking about IConfigureOptions?

jchannon:
Actually I can't pull ILogger out in the AddBotwin method here

davidfowl:
I'm saying remove the logging from AddBotwin

jchannon:
@davidfowl have updated the signature slightly but kept the logging in AddBotwin as this is where stuff is found to wire Botwin up and users want to know if their module 3 levels of transient dependency has been found.

src/BotwinExtensions.cs
```diff
@@ -4,36 +4,32 @@ namespace Botwin
...
         /// Adds Botwin to the specified <see cref="IApplicationBuilder"/>.
         /// </summary>
         /// <param name="builder">The <see cref="IApplicationBuilder"/> to configure.</param>
         /// <param name="options">A <see cref="BotwinOptions"/> instance.</param>
         /// <returns>A reference to this instance after the operation has completed.</returns>
-        public static IApplicationBuilder UseBotwin(this IApplicationBuilder builder, BotwinOptions options)
+        public static IApplicationBuilder UseBotwin(this IApplicationBuilder builder, BotwinOptions options = null)
         {
-            ApplyGlobalBeforeHook(builder, options);
+            var loggerFactory = builder.ApplicationServices.GetService<ILoggerFactory>();
```
poke: `loggerFactory` will never be `null` here, so no need to have any of the following null-coalescing calls.

src/BotwinExtensions.cs
```diff
@@ -4,36 +4,32 @@ namespace Botwin
...
+
+            var logger = loggerFactory?.CreateLogger(typeof(BotwinExtensions));
+
+            logger?.LogTrace("Adding Botwin to application");
```
poke: That’s not really useful. `app.UseX` adds middleware for `X`. There’s really no benefit to logging that.

src/BotwinExtensions.cs
```diff
@@ -44,19 +40,24 @@ public static IApplicationBuilder UseBotwin(this IApplicationBuilder builder, Bo
                 foreach (var module in scope.ServiceProvider.GetServices<BotwinModule>())
                 {
                     var moduleType = module.GetType();
-                    var distinctPaths = module.Routes.Keys.Select(route => route.path).Distinct();
+
+                    var distinctPaths = module.Routes.Keys.Select(route =>
+                    {
+                        logger?.LogTrace($"Found and adding {moduleType.Name} with {route.verb} /{route.path} to routing");
```
poke: This log should be done with the `MapRoute` call below. Collecting distinct paths does actually nothing about the route. It’s the `MapRoute` that does what this log says what happens.

Also, use structured logging instead of string interpolation.

src/BotwinExtensions.cs
```diff
@@ -44,19 +40,24 @@ public static IApplicationBuilder UseBotwin(this IApplicationBuilder builder, Bo
...
+                        routeBuilder.MapRoute(path, CreateRouteHandler(path, moduleType, logger));
                     }
                 }
             }
 
             return builder.UseRouter(routeBuilder.Build());
         }
 
-        private static RequestDelegate CreateRouteHandler(string path, Type moduleType)
+        private static RequestDelegate CreateRouteHandler(string path, Type moduleType, ILogger logger)
```
poke: I honestly don’t think you should use the same logger here. `CreateRouteHandler` basically creates a middleware. That middleware is run as part of the request pipeline. So it runs much later than any of the other things that logger (the `BotwinExtensions` logger!) is logging.

You should probably request a new logger inside of the middleware. And you should probably request an `ILogger<moduleType>` there so that people can actually figure out what module runs.

src/BotwinExtensions.cs
```diff
@@ -82,17 +83,20 @@ private static RequestDelegate CreateRouteHandler(string path, Type moduleType)
 
                 if (module.Before != null)
                 {
+                    logger?.LogTrace("Executing module before hook");
```
poke: Imho, if people want to be notified of their before/after handler being run, they should add logging calls inside of their modules themselves.

(same below with the after handler)

src/BotwinExtensions.cs
```diff
@@ -82,17 +83,20 @@ private static RequestDelegate CreateRouteHandler(string path, Type moduleType)
...
                     shouldContinue = await module.Before(ctx);
                 }
 
                 if (shouldContinue)
                 {
                     // run the route handler
+                    logger?.LogTrace($"Executing module route handler for {ctx.Request.Method} {path}");
```
poke: Use structured logging instead of string interpolation:

```
logger.LogTrace("Executing module route handler for {Method} {Path}", ctx.Request.Method, path);
```

src/BotwinExtensions.cs
```diff
@@ -115,42 +120,68 @@ private static RequestDelegate CreateRouteHandler(string path, Type moduleType)
...
-        private static void ApplyGlobalAfterHook(IApplicationBuilder builder, BotwinOptions options)
+        private static void ApplyGlobalAfterHook(IApplicationBuilder builder, BotwinOptions options, ILogger logger)
         {
             if (options?.After != null)
             {
+                logger?.LogTrace("Applying global after hook");
                 builder.Use(async (ctx, next) =>
                 {
                     await next();
+                    logger?.LogTrace("Executing global after hook");
```
poke: This again is inside of a middleware so it should probably use a different logger.

src/BotwinExtensions.cs
```diff
@@ -115,42 +120,68 @@ private static RequestDelegate CreateRouteHandler(string path, Type moduleType)
...
                     await options.After(ctx);
                 });
             }
+            else
+            {
+                logger?.LogTrace("No global after hook found");
```
poke: Absense of the “applying global after hook” log should be enough, no?

src/BotwinExtensions.cs
```diff
@@ -176,19 +208,22 @@ public static void AddBotwin(this IServiceCollection services)
 
             foreach (var module in modules)
             {
+                logger?.LogTrace($"Found {module.FullName}");
```
poke: These are pretty noisy. Maybe collect the module names, and then only log once with a list of detected modules?

Also: Use structured logging instead of string interpolation.

poke:
Yup, that looks much better.

You should still switch those interpolated strings to structured logging though.

poke:
The internal formatter for log messages is cached by the log format.

So a `logger.LogTrace($"Found {module.FullName}")` will create a new formatter and cache that for every module it finds. Structured formatting with `logger.LogTrace("Found {ModuleName}", module.FullName)` will only use a single formatter.

In addition, you have structured logging, so the module name being a `ModuleName` is also kept as separate information and log viewers are able to extract that information and filter by it. In general, you really should use proper structured logging everywhere, especially so in library code.

jchannon:
Woohoo! Thanks to all in reviewing this PR ❤️

davidfowl:
Passing the logging is still meh.

jchannon:
I don't know what the alternative is to log in the AddBotwin method

PR #88, opened 2018-03-08
Added optional DependencyContextAssemblyCatalog parameter in AddBotwin extension method

modernist:
Since the optional assembly parameter was removed from the `AddBotwin` extension method it was not possible to define additional assemblies for searching module and validator types and registering them in the `IServices` collection. This PR adds this capability and the dangling documentation for the `assemblies` parameter.

### DIFF
--- a/src/BotwinExtensions.cs
+++ b/src/BotwinExtensions.cs
@@ -146,10 +146,10 @@ namespace Botwin
         /// Adds Botwin to the specified <see cref="IServiceCollection"/>.
         /// </summary>
         /// <param name="services">The <see cref="IServiceCollection"/> to add Botwin to.</param>
-        /// <param name="assemblies">Optional array of <see cref="Assembly"/> to add to the services collection. If assemblies are not provided, Assembly.GetEntryAssembly is called.</param>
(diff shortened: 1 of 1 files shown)


jchannon:
Thanks.

Have you debugged DependencyContextAssemblyCatalog to see if it picks up the assemblies you want it to include.  If it's not picking your assemblies up I would say there is a bug in it as it should find all assemblies that are referencing Botwin

modernist:
I will test it more thoroughly, although my initial tests seem to pass. Another option would be to change the DependencyContextAssemblyCatalog constructor to accept an array of assemblies.

Is there another way to use modules and validators from other assemblies that I am not aware of? This is in essence what I'm trying to achieve. Up to version 3.1.0 I would put these in a different assembly and pass it in `AddBotwin`.

jchannon:
Nope that's how you'd use it.  If you look in the samples folder in this
repo you'll see a ValidatorOnlyProject which has a reference to
FluentValidation only but the main sample app references
ValidatorOnlyProject and this seems to work.

Could you illustrate what your project structure is like?

jchannon:
👀

jchannon:
Closed via #136

PR #90, opened 2018-03-22
Add FromStream extension

### LINKED ISSUE #89: Unable to respond with a Stream
I'm unable to work out how to reply with a streamed response using Botwin. Reading up on ASP.NET Core it looks liked I'd be able to do the following:

```csharp
[HttpGet]
public FileStreamResult GetTest()
{
  var stream = new MemoryStream(Encoding.ASCII.GetBytes("Hello World"));
  return new FileStreamResult(stream, new MediaTypeHeaderValue("text/plain"))
  {
    FileDownloadName = "test.txt"
  };
}
```
(https://stackoverflow.com/questions/42771409/how-to-stream-with-asp-net-core#42772150)

However, I can't work out how I can do this using the `HttpResponse` I have from the BotwinModule handler? 

Here's a snippet of my module, but it doesn't return any data:

```csharp
var models = this.journalReader.Read(start, end, type, ct);

var fileStream =
    new FileStreamResult(CreateExportStream(models, requestLog, ct), "application/csv")
        {
            FileDownloadName
                = $"journal-{type}.csv"
        };

res.Body = fileStream.FileStream;
```

jchannon:
Fixes #89

### DIFF
--- a/src/Response/ResponseExtensions.cs
+++ b/src/Response/ResponseExtensions.cs
@@ -1,7 +1,9 @@
 namespace Botwin.Response
 {
+    using System;
     using System.IO;
     using System.Linq;
+    using System.Net;
     using System.Net.Mime;
     using System.Threading;
     using System.Threading.Tasks;
@@ -69,17 +71,110 @@ namespace Botwin.Response
         /// <param name="stream">The <see cref="Stream"/> to copy from</param>
         /// <param name="contentType">The content type for the response</param>
         /// <param name="contentDisposition">The content disposition to allow file downloads</param>
-        /// <returns></returns>
-        public static async Task FromStream(this HttpResponse response, Stream stream, string contentType, ContentDisposition contentDisposition = null)
+        /// <returns><<see cref="Task"/>/returns>
+        public static async Task FromStream(this HttpResponse response, Stream source, string contentType, ContentDisposition contentDisposition = null)
         {
-            if (contentDisposition != null)
-            {
-                response.Headers.Add("Content-Disposition", contentDisposition.ToString());
-            }
+            long rangeStart;
+            long rangeEnd;
+            var contentLength = source.Length;
 
+            response.Headers["AcceptRanges"] = "bytes";
             response.ContentType = contentType;
 
-            await StreamCopyOperation.CopyToAsync(stream, response.Body, new long?(), 65536, response.HttpContext.RequestAborted);
+            if (contentDisposition != null)
+            {
+                response.Headers["Content-Disposition"] = contentDisposition.ToString();
+            }
+
+            // rangeHeader should be of the format "bytes=0-" or "bytes=0-12345" or "bytes=123-456"
+            var rangeHeader = response.HttpContext.Request.Headers["Range"].FirstOrDefault();
+
+            if (!string.IsNullOrEmpty(rangeHeader) && rangeHeader.Contains("="))
+            {
+                var rangeParts = rangeHeader.SplitOnFirst("=")[1].SplitOnFirst("-");
+                rangeStart = long.Parse(rangeParts[0]);
+                rangeEnd = rangeParts.Length == 2 && !string.IsNullOrEmpty(rangeParts[1])
+                    ? int.Parse(rangeParts[1]) // the client requested a chunk
+                    : contentLength - 1;
+
+                if (rangeStart < 0 || rangeEnd > contentLength - 1)
+                {
+                    response.StatusCode = (int)HttpStatusCode.RequestedRangeNotSatisfiable;
+                    await response.WriteAsync(string.Empty);
+                }
+                else
+                {
+                    response.Headers["ContentRange"] = $"bytes {rangeStart}-{rangeEnd}/{contentLength}";
+                    response.StatusCode = (int)HttpStatusCode.PartialContent;
+                    await GetResponseBodyDelegate(rangeStart, rangeEnd, contentLength, source, response.Body, response.HttpContext.RequestAborted);
+                }
+            }
+            else
+            {
+                rangeStart = 0;
+                rangeEnd = contentLength - 1;
+                await GetResponseBodyDelegate(rangeStart, rangeEnd, contentLength, source, response.Body, response.HttpContext.RequestAborted);
+            }
+        }
+
+        private static async Task GetResponseBodyDelegate(long rangeStart, long rangeEnd, long contentLength, Stream source, Stream destination, CancellationToken cancellationToken)
+        {
+            if (rangeStart == 0 && rangeEnd == contentLength - 1)
+            {
+                await StreamCopyOperation.CopyToAsync(source, destination, new long?(), 65536, cancellationToken);
+            }
+            else
+            {
+                if (!source.CanSeek)
+                {
+                    throw new InvalidOperationException("Sending Range Responses requires a seekable stream eg. FileStream or MemoryStream");
+                }
+
+                var totalBytesToSend = rangeEnd - rangeStart + 1;
+                const int BufferSize = 0x1000;
+                var buffer = new byte[BufferSize];
+                var bytesRemaining = totalBytesToSend;
+
+                source.Seek(rangeStart, SeekOrigin.Begin);
+                while (bytesRemaining > 0)
+                {
+                    var count = bytesRemaining <= buffer.Length
+                        ? await source.ReadAsync(buffer, 0, (int)Math.Min(bytesRemaining, int.MaxValue), cancellationToken)
+                        : await source.ReadAsync(buffer, 0, buffer.Length, cancellationToken);
+
+                    try
+                    {
+                        await destination.WriteAsync(buffer, 0, count, cancellationToken);
+                        await destination.FlushAsync(cancellationToken);
+                        bytesRemaining -= count;
+                    }
+                    catch (Exception httpException)
+                    {
+                        /* in Asp.Net we can call HttpResponseBase.IsClientConnected
+                        * to see if the client broke off the connection
+                        * and avoid trying to flush the response stream.
+                        * instead I'll swallow the exception that IIS throws in this situation
+                        * and rethrow anything else.*/
+                        if (httpException.Message
+                            == "An error occurred while communicating with the remote host. The error code is 0x80070057.") return;
+
+                        throw;
+                    }
+                }
+            }
+        }
+
+        public static string[] SplitOnFirst(this string strVal, string needle)
+        {
+            if (strVal == null)
+            {
+                return new string[0];
+            }
+
+            var pos = strVal.IndexOf(needle, StringComparison.OrdinalIgnoreCase);
+            return pos == -1
+                ? new[] { strVal }
+                : new[] { strVal.Substring(0, pos), strVal.Substring(pos + needle.Length) };
         }
     }
 }
--- a/src/Response/ResponseExtensions.cs
+++ b/src/Response/ResponseExtensions.cs
@@ -78,7 +78,7 @@ namespace Botwin.Response
             long rangeEnd;
             var contentLength = source.Length;
 
-            response.Headers["AcceptRanges"] = "bytes";
+            response.Headers["Accept-Ranges"] = "bytes";
             response.ContentType = contentType;
 
             if (contentDisposition != null)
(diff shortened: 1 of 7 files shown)


src/Response/ResponseExtensions.cs
```diff
@@ -58,5 +63,118 @@ public static class ResponseExtensions
...
+                        if (httpException.Message
+                            == "An error occurred while communicating with the remote host. The error code is 0x80070057.") return;
+
+                        throw;
+                    }
+                }
+            }
+        }
+
+        public static string[] SplitOnFirst(this string strVal, string needle)
```
poke: A public extension on `string`? This should probably be private or internal or something.

src/Response/ResponseExtensions.cs
```diff
@@ -58,5 +63,118 @@ public static class ResponseExtensions
...
+        {
+            if (strVal == null)
+            {
+                return new string[0];
```
poke: `Array.Empty<string>()`

src/Response/ResponseExtensions.cs
```diff
@@ -58,5 +63,118 @@ public static class ResponseExtensions
...
+            }
+
+            var pos = strVal.IndexOf(needle, StringComparison.OrdinalIgnoreCase);
+            return pos == -1
+                ? new[] { strVal }
+                : new[] { strVal.Substring(0, pos), strVal.Substring(pos + needle.Length) };
```
poke: Isn’t this pretty much the same as `strVal.Split(new [] { needle }, 2, StringSplitOptions.RemoveEmptyEntries)`?

src/Response/ResponseExtensions.cs
```diff
@@ -58,5 +63,118 @@ public static class ResponseExtensions
...
+
+            // rangeHeader should be of the format "bytes=0-" or "bytes=0-12345" or "bytes=123-456"
+            var rangeHeader = response.HttpContext.Request.Headers["Range"].FirstOrDefault();
+
+            if (!string.IsNullOrEmpty(rangeHeader) && rangeHeader.Contains("="))
+            {
+                var rangeParts = rangeHeader.SplitOnFirst("=")[1].SplitOnFirst("-");
+                rangeStart = long.Parse(rangeParts[0]);
+                rangeEnd = rangeParts.Length == 2 && !string.IsNullOrEmpty(rangeParts[1])
+                    ? int.Parse(rangeParts[1]) // the client requested a chunk
```
poke: `rangeParts[0]` is parsed as `long`, while `rangeParts[1]` is parsed as `int`. Also, maybe use `TryParse` given that the `rangeHeader` format is not strictly validated.

src/Response/ResponseExtensions.cs
```diff
@@ -58,5 +63,118 @@ public static class ResponseExtensions
...
+                    : contentLength - 1;
+
+                if (rangeStart < 0 || rangeEnd > contentLength - 1)
+                {
+                    response.StatusCode = (int)HttpStatusCode.RequestedRangeNotSatisfiable;
+                    await response.WriteAsync(string.Empty);
```
poke: Do you actually have to write an empty string here? Can’t you just do nothing?

src/Response/ResponseExtensions.cs
```diff
@@ -58,5 +63,118 @@ public static class ResponseExtensions
...
+
+        /// <summary>
+        /// Copy a stream into the response body
+        /// </summary>
+        /// <param name="response">Current <see cref="HttpResponse"/></param>
+        /// <param name="stream">The <see cref="Stream"/> to copy from</param>
+        /// <param name="contentType">The content type for the response</param>
+        /// <param name="contentDisposition">The content disposition to allow file downloads</param>
+        /// <returns><<see cref="Task"/>/returns>
+        public static async Task FromStream(this HttpResponse response, Stream source, string contentType, ContentDisposition contentDisposition = null)
```
poke: nit: `FromStreamAsync`

src/Response/ResponseExtensions.cs
```diff
@@ -58,5 +63,118 @@ public static class ResponseExtensions
...
+            }
+            else
+            {
+                rangeStart = 0;
+                rangeEnd = contentLength - 1;
+                await GetResponseBodyDelegate(rangeStart, rangeEnd, contentLength, source, response.Body, response.HttpContext.RequestAborted);
+            }
+        }
+
+        private static async Task GetResponseBodyDelegate(long rangeStart, long rangeEnd, long contentLength, Stream source, Stream destination, CancellationToken cancellationToken)
```
poke: nit: `GetResponseBodyDelegateAsync`.

src/Response/ResponseExtensions.cs
```diff
@@ -58,5 +63,118 @@ public static class ResponseExtensions
...
+        {
+            if (rangeStart == 0 && rangeEnd == contentLength - 1)
+            {
+                await StreamCopyOperation.CopyToAsync(source, destination, new long?(), 65536, cancellationToken);
```
poke: Maybe make the `rangeStart` and `rangeEnd` nullable to basically disable the splitting?

Also, maybe return here, so you don’t need to indent the whole rest of the method?

src/Response/ResponseExtensions.cs
```diff
@@ -58,5 +63,118 @@ public static class ResponseExtensions
...
+            }
+            else
+            {
+                if (!source.CanSeek)
+                {
+                    throw new InvalidOperationException("Sending Range Responses requires a seekable stream eg. FileStream or MemoryStream");
+                }
+
+                var totalBytesToSend = rangeEnd - rangeStart + 1;
+                const int BufferSize = 0x1000;
```
poke: Either move this constant out, or inline it since it’s just used a single time?

src/Response/ResponseExtensions.cs
```diff
@@ -58,5 +63,118 @@ public static class ResponseExtensions
...
+                var buffer = new byte[BufferSize];
+                var bytesRemaining = totalBytesToSend;
+
+                source.Seek(rangeStart, SeekOrigin.Begin);
+                while (bytesRemaining > 0)
+                {
+                    var count = bytesRemaining <= buffer.Length
+                        ? await source.ReadAsync(buffer, 0, (int)Math.Min(bytesRemaining, int.MaxValue), cancellationToken)
```
poke: `Math.Min(bytesRemaining, int.MaxValue)` should never return something other than `bytesRemaining`.

You probably meant to do this to avoid the ternary?
```
var count = await source.ReadAsync(buffer, 0, (int)Math.Min(bytesRemaining, buffer.Length), cancellationToken);
```

test/Botwin.Tests/ResponseFromStreamTests.cs
```diff
@@ -0,0 +1,68 @@
...
+
+            //Then
+            Assert.Equal("hi", body);
+        }
+        
+        [Fact]
+        public async Task Should_set_content_disposition_header_if_supplied()
+        {
+            //Given & When
+            var response = await this.httpClient.GetAsync("/downloadwithcd");
```
poke: Is there a reason you have three separate tests for the same method/parameter combination? Why not just test all three things at once?

test/Botwin.Tests/ResponseFromStreamTests.cs
```diff
@@ -0,0 +1,68 @@
...
+        }
+        
+        [Fact]
+        public async Task Should_not_set_content_disposition_header_by_default()
+        {
+            //Given & When
+            var response = await this.httpClient.GetAsync("/download");
+
+            //Then
+            Assert.Null(response.Content.Headers.ContentDisposition);
```
poke: Should probably assert, that the rest of the response is still correct.

jchannon:
Sorry @danbarua removed your code now 😄 but now you can use Botwin and not have to use your own extensions! Feel free to review 😛 

@poke ready to roll I think!?

src/Response/ResponseExtensions.cs
```diff
@@ -58,5 +63,59 @@ public static class ResponseExtensions
...
+            {
+                response.Headers["Content-Disposition"] = contentDisposition.ToString();
+            }
+
+            // rangeHeader should be of the format "bytes=0-" or "bytes=0-12345" or "bytes=123-456"
+            var rangeHeader = response.HttpContext.Request.GetTypedHeaders().Range;
+
+            if (rangeHeader != null)
+            {
+                //Server should return multipart/byteranges; if asking for more than one range but pfft...
```
danbarua: Ain't nobody got time for that!

src/Response/ResponseExtensions.cs
```diff
@@ -58,5 +63,59 @@ public static class ResponseExtensions
...
+                rangeStart = rangeHeader.Ranges.First().From;
+                rangeEnd = rangeHeader.Ranges.First().To ?? contentLength -1;
```
poke: nit: space after the `-`.

Also: Is there a reason why the variables `rangeStart` and `rangeEnd` are defined at the beginning of the method?

Also also: Can `Ranges.First()` fail?

src/Response/ResponseExtensions.cs
```diff
@@ -58,5 +63,59 @@ public static class ResponseExtensions
...
+                        throw new InvalidOperationException("Sending Range Responses requires a seekable stream eg. FileStream or MemoryStream");
+                    }
+                    
+                    source.Seek(rangeStart.Value, SeekOrigin.Begin);
+                    await StreamCopyOperation.CopyToAsync(source, response.Body, rangeEnd.Value - rangeStart.Value + 1, 65536, response.HttpContext.RequestAborted);
+                }
+            }
+            else
+            {
+                await StreamCopyOperation.CopyToAsync(source, response.Body, new long?(), 65536, response.HttpContext.RequestAborted);
```
poke: `new long?()`? Use `default(long?)` or just the `default` keyword, or just `null`.

test/Botwin.Tests/ResponseFromStreamTests.cs
```diff
@@ -0,0 +1,88 @@
...
+        {
+            //Given & When
+            var response = await this.httpClient.GetAsync("/downloadwithcd");
+            var body = await response.Content.ReadAsStringAsync();
+            var filename = response.Content.Headers.ContentDisposition.FileName;
+
+            //Then
+            Assert.Equal("application/csv", response.Content.Headers.ContentType.MediaType);
+            Assert.Equal("hi", body);
+            Assert.Equal("bytes", response.Headers.AcceptRanges.FirstOrDefault());
```
poke: nit: Consistency: `response.Headers` vs. `response.Content.Headers`

src/Response/ResponseExtensions.cs
```diff
@@ -58,5 +63,59 @@ public static class ResponseExtensions
...
         }
+
+        /// <summary>
+        /// Copy a stream into the response body
+        /// </summary>
+        /// <param name="response">Current <see cref="HttpResponse"/></param>
+        /// <param name="stream">The <see cref="Stream"/> to copy from</param>
+        /// <param name="contentType">The content type for the response</param>
+        /// <param name="contentDisposition">The content disposition to allow file downloads</param>
+        /// <returns><<see cref="Task"/>/returns>
```
poke: Invalid XML: Move the `<see …/>` one character to the left (double `<<` and no `<` before the `/returns`)

aarondandy: > A task representing the asynchronous copy to the response body.

src/Response/ResponseExtensions.cs
aarondandy: I played around with `RangeHeaderValue` and don't think it will parse without at least one range given.

src/Response/ResponseExtensions.cs
aarondandy: I feel like it should at least throw not supported. Would it not just be as easy as a little `foreach(var range in rangeHeader.Ranges)` ?

aarondandy: Nope, looks hard. but...
```csharp
if (rangeHeader.Ranges.Count != 1)
{
    throw something;
}
```
could prevent a surprise.

jchannon:
Things like 3-1 are impossible to test as the framework won't allow it as it's invalid values so trying to assign headers with dodgy values fails.  In live tests if you parse that in it will fail on the `var rangeHeader = response.HttpContext.Request.GetTypedHeaders().Range;` and return null and therefore return the whole stream

src/Response/ResponseExtensions.cs
jchannon: As @aarondandy says. Seems to be a lot of validation under the hood

test/Botwin.Tests/ResponseFromStreamTests.cs
jchannon: Blame ASPNET team, some are exposed on Headers whilst others are on response.Content.Headers. Weird

src/Response/ResponseExtensions.cs
jchannon: if multiple are parsed then they will get the first.

test/Botwin.Tests/ResponseFromStreamTests.cs
poke: `HttpContentHeaders` and `HttpResponseHeaders`… meh >_<"

jchannon:
Ah I didn't know about `TryAddWithoutValidation `. My colombo skills are weak!

New tests added.

Why are the response headers so inconsistent. Just shove everything under response.Headers. Much simpler

jchannon:
Thanks to all that reviewed this! ❤️ 👍

PR #96, opened 2018-04-07
Alternative logging approach

poke:
This is a follow-up to #87, with an alternative attempt to solve this so that people don’t have to pass an actual logger to it. This basically delays the logging to the `UseBotwin()` call and stores the results in a diagnostics object.

Another alternative would have been to resolve the types once at the beginning of `UseBotwin`, that way we could avoid that temporary object but we also would have to create every object once for nothing (as there’s unfortunately no way to see what’s registered with a service provider except resolving it).

Of course, the third alternative would be to change how Botwin is resolving the modules, similar to how MVC resolves controllers when they are not registered through DI. But that would have been a bit more complicated.

### DIFF
--- a/src/CarterExtensions.cs
+++ b/src/CarterExtensions.cs
@@ -21,6 +21,13 @@ namespace Carter
         /// <returns>A reference to this instance after the operation has completed.</returns>
         public static IApplicationBuilder UseCarter(this IApplicationBuilder builder, CarterOptions options = null)
         {
+            var diagnostics = builder.ApplicationServices.GetService<CarterDiagnostics>();
+            if (diagnostics != null)
+            {
+                var logger = builder.ApplicationServices.GetService<ILoggerFactory>().CreateLogger(typeof(CarterDiagnostics));
+                diagnostics.LogDiscoveredModules(logger);
+            }
+
             ApplyGlobalBeforeHook(builder, options);
 
             ApplyGlobalAfterHook(builder, options);
@@ -133,8 +140,8 @@ namespace Carter
                     var loggerFactory = ctx.RequestServices.GetService<ILoggerFactory>();
                     var logger = loggerFactory.CreateLogger("Carter.GlobalBeforeHook");
                     logger.LogTrace("Executing global before hook");
-                    
-                    var carryOn = await options.Before(ctx); 
+
+                    var carryOn = await options.Before(ctx);
                     if (carryOn)
                     {
                         logger.LogTrace("Executing next handler after global before hook");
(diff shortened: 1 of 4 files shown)


jchannon:
Looks good cc @davidfowl

Would be interested to see how MVC resolves stuff. Botwin currently scans assemblies and registers them

src/BotwinExtensions.cs
```diff
@@ -21,6 +21,13 @@ public static class BotwinExtensions
         /// <returns>A reference to this instance after the operation has completed.</returns>
         public static IApplicationBuilder UseBotwin(this IApplicationBuilder builder, BotwinOptions options = null)
         {
+            var diagnostics = builder.ApplicationServices.GetService<BotwinDiagnostics>();
+            if (diagnostics != null)
+            {
+                var logger = builder.ApplicationServices.GetService<ILoggerFactory>().CreateLogger(typeof(BotwinDiagnostics));
+                diagnostics.LogDiscoveredModules(logger);
```
jchannon: Rather than adding types to lists would it be more performant to just resolve modules, validators etc and log them here? I don't really like that as it seems a perf hit.

I know the api currently is a bit yucky but its a bit more performant

poke: As mentioned above, this would have been an alternative implementation.

Collecting the types in a list will only require some additional memory, although I would not expect `Type` references to be heavy.

On the other hand, resolving all types once at the beginning will actively create those objects, and all their dependencies. Since there are no restrictions on what users could do and what things they could depend on, I would expect this to have a bigger impact. And there’s a possibility of side effects too. Not to mention that resolving objects just to throw them away feel really bad 🙈 

As far as performance goes, this should not really make any difference, as this only ever runs once when the application starts up. So this would only affect the startup time (although I wouldn’t expect this to be noticeable).

samples/BotwinSample/Startup.cs
```diff
@@ -7,22 +7,14 @@ namespace Botwin.Samples
...
-        {
-            this.loggerFactory = loggerFactory;
-        }
         public void ConfigureServices(IServiceCollection services)
         {
             services.AddSingleton<IActorProvider, ActorProvider>();
             
-            //If you want to log what Botwin finds you need to pass in a ILoggerFactory as ASP.Net Core does not offer a clean way to log extensions to IServiceCollection
-            services.AddBotwin(this.loggerFactory);
+            services.AddBotwin(enableDiagnostics: true);
```
davidfowl: I would make this a flag on an Options type. It should look like this:

```C#
services.AddBotwin(options =>
{
    options.EnableDiagnostics = true;
});
```

Plus it sets up Botwin for adding future options.

davidfowl:
I prefer this approach.

> Would be interested to see how MVC resolves stuff. Botwin currently scans assemblies and registers them

MVC doesn't log during ConfigureServices (because it can't).

samples/BotwinSample/Startup.cs
poke: Yeah, I wanted to do that too, but then I would have to call the action myself once within the `AddBotwin` to figure out whether I need to collect the types or not. And I couldn’t take the options from other sources then either.

So this would only work properly if the type collection would always be enabled and the flag just controls whether to do the logging within the `UseBotwin()` call.

jchannon:
Needs a rebase 😄

poke:
🙄

src/CarterExtensions.cs
```diff
@@ -148,33 +155,27 @@ private static void ApplyGlobalBeforeHook(IApplicationBuilder builder, CarterOpt
...
-
-        /// <summary>
-        /// Adds Carter to the specified <see cref="IServiceCollection"/>.
-        /// </summary>
-        /// <param name="services">The <see cref="IServiceCollection"/> to add Carter to.</param>
-        /// <param name="assemblies">Optional array of <see cref="Assembly"/> to add to the services collection. If assemblies are not provided, Assembly.GetEntryAssembly is called.</param>
-        /// <param name="logger">Optional <see cref="ILogger"/> to be passed for tracing of Carter setup</param>
-        public static void AddCarter(this IServiceCollection services, ILogger logger = null)
+        /// <param name="enableDiagnostics">Whether to enable diagnostics at startup.</param>
+        public static void AddCarter(this IServiceCollection services, bool enableDiagnostics = false)
```
jchannon: I wonder if we remove the `enableDiagnostics` and just rely on the LogLevel, therefore we always add the CarterDiagnostics to the IOC container and call `LogDiscoveredModules` but depending on the log level it will/ill not get logged. Thoughts?

jchannon:
Does SetMinimumLevel need removing now you added the log level in the json file? https://github.com/CarterCommunity/Carter/blob/master/samples/CarterSample/Program.cs#L13

jchannon:
Is it possible to detect the logging level, if so how about we change the logging to Debug rather than Trace and remove the `enableDiagnostics` and do an `if` on the logging level and if Debug register and log

poke:
Detecting the active log level requires a logger.

jchannon:
Ah balls that's what you meant earlier with `logger.IsEnabled`

jchannon:
Thanks!! 👍

PR #99, opened 2018-04-08
Diagnostics permanently turned on

jchannon:
Change to LogDebug for diagnostics and always have it turned on and let loglevel determine if it appears in logs


jchannon:
Ping @poke

jchannon:
Thanks!

PR #100, opened 2018-04-09
Add rd.xml to csproj

markrendle:
OK, this gets past the problem on Twitter and now hits a NRE, but within Carter code.

### DIFF
--- a/samples/NativeCarter/CarterNativeSample.csproj
+++ b/samples/NativeCarter/CarterNativeSample.csproj
@@ -6,6 +6,8 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore" Version="2.0.0" />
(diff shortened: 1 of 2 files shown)


samples/NativeCarter/CarterNativeSample.csproj
```diff
@@ -6,6 +6,8 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore" Version="2.0.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.Core" Version="2.0.1" />
```
jchannon: Any ideas why this fixes the problem?  I know its a sample app but Carter apps shouldn't depend on MVC

jchannon:
Thanks 👍 👍

PR #102, opened 2018-04-23
Appveyor CI

mr-sandy:
Added .appveyor configuration to run `dotnet test` and `dotnet pack` on each commit to `master`.

The resulting nuget package will be available on the Appveyor project feed.

I have alternative config for Appveyor Linux beta, if that's more desirable.

I've done nothing to resolve the hard-coding of the base version number within the .appveyor config.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -21,7 +21,9 @@ build_script:
 test_script:
   - cmd: dotnet test test\Carter.Tests\Carter.Tests.csproj
 
+nuget:
+  disable_publish_on_pr: true
+  
 artifacts:
   - path: '**\*.nupkg'
     name: packages
-
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,3 +1,5 @@
+image: ubuntu
+
 version: 3.6.0-{build}
 
 branches:
@@ -15,11 +17,10 @@ dotnet_csproj:
   file_version: '{version}'
   informational_version: '{version}'
 
-build_script:
-  - cmd: dotnet pack src\Carter.csproj -c Release -o nupkgs
+configuration: Release
 
-test_script:
-  - cmd: dotnet test test\Carter.Tests\Carter.Tests.csproj
+build:
+  verbosity: minimal
 
 nuget:
   disable_publish_on_pr: true
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,27 @@
+version: 3.6.0-{build}
+
+branches:
+  only:
+    - master
+
+max_jobs: 1
+
+dotnet_csproj:
+  patch: true
+  file: '**\*.csproj'
+  version: '{version}'
+  package_version: '{version}'
+  assembly_version: '{version}'
+  file_version: '{version}'
+  informational_version: '{version}'
+
+build_script:
+  - cmd: dotnet pack src\Carter.csproj -c Release -o nupkgs
+
+test_script:
+  - cmd: dotnet test test\Carter.Tests\Carter.Tests.csproj
+
+artifacts:
+  - path: '**\*.nupkg'
+    name: packages
+
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,4 @@
-version: 3.6.0-{build}
+version: 3.6.0.{build}
 
 branches:
   only:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,4 @@
-version: 3.6.0.{build}
+version: 3.6.0-{build}
 
 branches:
   only:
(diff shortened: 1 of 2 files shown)


jchannon:
I think we'd need to have it run on pull requests to make sure no-one has broken anything, just need to look into disabling packaging per PR and if we can't is it a problem

jchannon:
Add this 

```

nuget:
  disable_publish_on_pr: true
```

mr-sandy:
> To avoid packaging on PR builds, the easiest thing to do is to disable pull requests inside the project's webhook configuration for Appveyor.

Seems I might have identified the _second_ easiest technique.  

Have now added...

```
nuget:
  disable_publish_on_pr: true
```

as suggested.

jchannon:
Cool. Do you need a windows and linux file for it to run on both platforms?

mr-sandy:
The instruction on which build env/slave is encapsulated in the .appveyor config.  Since dotnet core is first class entity in their linux environment, you don't need to specify such explicit build or test commands.

You just need to choose if you'd rather use windows or linux.

Here's the other config...

```
image: ubuntu

version: 3.6.0-{build}

branches:
  only:
    - master

max_jobs: 1

dotnet_csproj:
  patch: true
  file: '**\*.csproj'
  version: '{version}'
  package_version: '{version}'
  assembly_version: '{version}'
  file_version: '{version}'
  informational_version: '{version}'

configuration: Release

build:
  verbosity: minimal

nuget:
  disable_publish_on_pr: true
  
artifacts:
  - path: '**\*.nupkg'
    name: packages

```

You do need to update your csprov to package on build.

jchannon:
But you can't have it run on two platforms?

mr-sandy:
You can't have it build on two platforms.  

The output of the build will be cross-platform.

Am I misunderstanding something?

jchannon:
Nope just surprised they dont allow you to run on a build on a linux and windows

mr-sandy:
You have to choose.

https://twitter.com/davidshrigley/status/796049886995169282

jchannon:
Lets go with linux then :)

jchannon:
although https://www.appveyor.com/docs/getting-started-with-appveyor-for-linux/#running-windows-and-linux-builds-side-by-side

JoeStead:
What would building on both platforms give? I do agree that it needs to build/test on PRs too though

appveyor.yml
```diff
@@ -0,0 +1,30 @@
+image: ubuntu
+
+version: 3.6.0-{build}
+
+branches:
```
jchannon: Will need to remove that branch check as we'll want it to run all branches

appveyor.yml
```diff
@@ -0,0 +1,30 @@
...
+
+dotnet_csproj:
+  patch: true
+  file: '**\*.csproj'
+  version: '{version}'
+  package_version: '{version}'
+  assembly_version: '{version}'
+  file_version: '{version}'
+  informational_version: '{version}'
+
```
jchannon: Need a `dotnet test` thing?

appveyor.yml
mr-sandy: Then it will package on all branches.

appveyor.yml
mr-sandy: It's picked up by convention

mr-sandy: Here's my last build output https://ci.appveyor.com/project/mr-sandy/carter/build/3.6.0-8

appveyor.yml
jchannon: nope `disable_publish_on_pr: true` :)

appveyor.yml
jchannon: interesting :)

appveyor.yml
mr-sandy: what will that do for feature branches pre-pr?  wouldn't they result in a build?

appveyor.yml
mr-sandy: :)

appveyor.yml
jchannon: yeah so appveyor will build and test but not pack which is what we want for PRs

jchannon:
Thanks! 👍

src/Carter.csproj
```diff
@@ -10,6 +10,7 @@
     <PackageLicenseUrl>https://github.com/CarterCommunity/Carter/blob/master/LICENSE</PackageLicenseUrl>
     <PackageIconUrl>https://github.com/CarterCommunity/Media/raw/master/carterlogo.png</PackageIconUrl>
     <LangVersion>latest</LangVersion>
+    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
```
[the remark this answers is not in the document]
mr-sandy: https://www.appveyor.com/docs/getting-started-with-appveyor-for-linux/

> AppVeyor provides automatic packaging for the following types of .NET Core projects:
> 
> Any .NET Core project with “Generate NuGet package on build” option enabled on “Package” tab enabled.
> ASP.NET Core web application.
> .NET Core console application.

Issue #115, opened 2018-06-07
Profile Carter


jchannon:
After the lovely @benaadams added Carter to the TechEmpower perf tests https://github.com/TechEmpower/FrameworkBenchmarks MVC was fractionally faster on plain text.

After reading the results at 9am they sent me down a nasty path of drug and alcohol abuse. I had an existential crisis and questioned my sanity.  I didn't listen to friends and ignored their advice.  Luckily I passed out the other side a better man. By 9.30am I was back ready to take on the beast.

So we need to run some profiling to see if we can see any hot spots in the code because in theory Carter should be faster, it's just a wrapper over RoutingMiddleware.

LET'S DO THIS!

jchannon:
My initial guess, and it is a guess is that we resolve status code handlers, the module, ILoggerFactory per request and that might be an issue when you're hitting 1m req/sec

chrissie1:
More alcohol and drugs.

PR #121, opened 2018-07-18
Optimized Route Handler

RagingKore:
This targets #115. I had a quick look and confirmed what you pointed out @jchannon.

So I moved module, logger and status code handlers resolution out of the route handler.

### DIFF
--- a/src/CarterExtensions.cs
+++ b/src/CarterExtensions.cs
@@ -1,6 +1,6 @@
 namespace Carter
 {
-    using System;
+    using System.Collections.Generic;
     using System.IO;
     using System.Linq;
     using System.Reflection;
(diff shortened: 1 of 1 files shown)


RagingKore:
I think the DefaultValidatorLocator could be optimized to resolve map all validators immediately instead of doing it per request and caching.

You not only have a faster first time execution of a request, but you wont need to throw and exception on runtime about having multiple validators for a type.

RagingKore:
I also have the new proposed DefaultValidatorLocator ready, but unfortunatly, because Carter currently has no way to manually register Validators it is not possible to test the exception.

jchannon:
Cheers 👍

Issue #112, opened 2018-05-24
Route Level security


jchannon:
Was thinking of something like this.

Thoughts?

`this.Get("/", async ctx => await this.RequiresAuthentication(async innerctx => await innerctx.WriteAsync("authed")));`

jchannon:
Pinging @RagingKore

JoeStead:
It's almost as if someone else has wrote this code :) I still don't like it for the record

jchannon:
I just made it better

JoeStead:
But broken ;)

Options I can see are:

* Make the route return something, so you can return the result of a requires auth method if necessary, and exit early
* Throw some kind of exception that gets converted into the correct response 🤢
* An optional pre-hook you can add to the routes individually, and return a boolean similar ot the global before hook
* Just work it out in the routes:

```
if(!IsAuthenticated)
{
  ctx.Response.StatusCode = 401;
   return;
}

//route code
```

That said, how often do you have a module where some routes will have different auth settings than others? There's nothing stopping you having multiple modules with the same base path if you wanted. I'm interested in usecase, mostly

jchannon:
Or...

```
            this.Get("/", async ctx =>
            {
                var authenticated = ctx?.User?.Identity != null && ctx.User.Identity.IsAuthenticated;

                if (!authenticated)
                {
                    ctx.Response.StatusCode = 401;
                    return false;
                }

                return true;
            }, async (request, response, arg3) => await response.WriteAsync("Authed"));
```

JoeStead:
You mean: 
> An optional pre-hook you can add to the routes individually, and return a boolean similar ot the global before hook 

:D

JoeStead:
Yeah, but you'd probably move that into a method so it would look more like 

```
this.Get("/", this.RequiresAuth, async (req, resp, route) => await response.WriteAsync("Authed"));` 
```

Otherwise there's no point using a pre-hook for it, and it should just be part of the route body, which I think should be the case anyway

jchannon:
Which has now led me to wonder if routes should be

`this.Get("/", async (req, resp, routeData, next) => {await response.WriteAsync("First"); await next()), async (req, resp, route) => await response.WriteAsync("Second"); await next())`

Basically middleware....

So first can be a security check and if not authed dont call `next()`

JoeStead:
[...]

PR #111, opened 2018-05-18
Carter diagnostics

JoeStead:
The stuff in the Diagnostics class / module is awful at the moment, but I thought I'd stick a WIP PR up for now :)

PR #122, opened 2018-07-18
[WIP] Removed assembly scanning by adding a bootstrapper

RagingKore:
Removed assembly scanning and 99% of DI dependency by adding a simple bootstrapper, that still needs to be registered since Validation and Response extensions will require access to validators and negotiators.

CarterOptions still contain the global hooks setup, but was considering moving those to the bootstraper too.

This also opens the door to delegate composition as we now depend on a CarterRoute, not the module.

Performance optimizations are included.

* Added CarterBootstrapper to help register and keep dependencies.
* Added CarterRoute to better flow the required information to the route builder.
* Refactored CarterModule to build the before/after pipeline immediately.
* Fixed extensions to now resolve the bootstrapper.
* Removed DefaultValidatorLocator and added required logic to bootstrapper.
* Refactored CarterDiagnostics to a static function.

This targets #115 and somewhat supersedes #116 and #121.


RagingKore:
Maybe it would be nice to keep backward compatibility by allowing a way to still run assembly scanning.

RagingKore:
This targets #111 and somewhat supersedes #116 and #121.

jchannon:
@thefringeninja fancy giving this some 👀

samples/CarterSample/Startup.cs
```diff
@@ -13,9 +17,20 @@ public class Startup
...
+                bootstrapper
+                    .RegisterModules(
+                        new FunctionalProgrammingModule(),
+                        new ActorsModule(new ActorProvider()),
+                        new CastMemberModule(),
+                        new HomeModule(),
+                        new ErrorModule())
+                    .RegisterValidators(new ActorValidator(), new DirectorValidator())
+                    .RegisterResponseNegotiators(new DefaultJsonResponseNegotiator())
+                    .RegisterStatusCodeHandlers(new ConflictStatusCodeHandler())
```
jasonmitchell: I really like being able to explicitly register this stuff 👍

src/CarterExtensions.cs
```diff
@@ -13,6 +14,66 @@ namespace Carter
...
+            services.AddSingleton<IEnumerable<IResponseNegotiator>>(bootstrapper.ResponseNegotiators);
+            services.AddSingleton<ICarterBootstrapper>(bootstrapper);
+            services.AddRouting();
+        }
+
+        /// <summary>
+        /// Adds Carter to the specified <see cref="IServiceCollection"/>.
+        /// </summary>
+        /// <param name="services">The <see cref="IServiceCollection"/> to add Carter to.</param>
+        /// <param name="configure">Allows <see cref="CarterBootstrapper"/> to be configured.</param>
```
jasonmitchell: Copy and paste comment?

src/CarterDependencyScanner.cs
```diff
@@ -0,0 +1,48 @@
...
+    {
+        private readonly DependencyContextAssemblyCatalog catalog;
+        
+        public CarterDependencyScanner() => this.catalog = new DependencyContextAssemblyCatalog();
+
+        public CarterDependencyScanner(Assembly assembly) => this.catalog = new DependencyContextAssemblyCatalog(assembly);
+
+        public IEnumerable<Type> ScanValidators() 
+            => this.catalog.GetAssemblies()
+                .SelectMany(ass => ass.GetTypes())
```
jasonmitchell: lol

src/CarterExtensions.cs
```diff
@@ -13,6 +14,66 @@ namespace Carter
...
+        public static void AddCarterScoped(this IServiceCollection services)
```
jasonmitchell: Why `AddCarterScoped` and not just `AddCarter`?  Not really a fan of this method name.

src/CarterRoute.cs
```diff
@@ -0,0 +1,24 @@
+namespace Carter
+{
+    using System;
+    using Microsoft.AspNetCore.Http;
+
+    public class CarterRoute
```
jasonmitchell: The introduction of this seems like it might be useful for implementing the `.Pipe(...)` functionality mentioned in #112 👍

src/CarterExtensions.cs
jchannon: Yup should be the same

jchannon:
I wonder if we should tackle the old "scoped" issue with a bootstrapper somehow so we could get rid of the `if (bootstrapper != null)` and maybe if the collections are empty on the bootstrapper we do the assembly scanning anyway??

RagingKore:
That was the idea, I just waited a bit to get feedback since it will still require a bit of work and I wanted to make sure that we can continue down this road.

src/CarterRoute.cs
RagingKore: I actually already have a version of the pipe stuff ready but I'm saving it for the next PR lol

jchannon:
We have CarterConfigurator which I think will fix the ass scanning

Issue #123, opened 2018-07-22
Register Validators as scoped instances rather than singleton


mysticmind:
I had a setup a validation as below wherein `RoleManager` via constructor injection is a scoped instance and this resulted in an error `Cannot consume scoped service in Singleton` (validators are registered as Singleton within `CarterExtensions.cs`).
```
public class CommandValidator : AbstractValidator<Command>
{
    public CommandValidator(RoleManager<UserRole> roleManager)
    {
        RuleFor(c => c.Name).CustomAsync(async (name, context, cancel) =>
        {
            if (string.IsNullOrEmpty(name))
            {
                context.AddFailure("Role name is required");
            }
            else
            {
                var role = await roleManager.FindByNameAsync(name);

                if (role != null)
                {
                    context.AddFailure("Role name already exists");
                }
            }
        });
    }
}
```

It would be good to register validators as scoped instances.

@jchannon if you could confirm on this change, I would be happy to send my first PR to Carter :-)

JoeStead:
It's probably best to make this configurable, in the cases where you can have validators as a singleton, that would be preferable because it will have an impact on performance. There have been other discussions around the use of the container / how things are registered, and I imagine those discussions will continue for a while :)

mysticmind:
@JoeStead understand your comments around requiring it to be configurable. Let me get my head around it and see if I can make it configurable. Apparently I did send a PR to change it to scoped, I think it can be ignored.

damianh:
@mysticmind Looks like you are doing a bit of CQRS there. You have a race condition in your validator and your domain logic is leaking - two concurrent requests with the same `name` will succeed the `FindByNameAsync()` but then you'll get duplicate / exception in next layer (domain or db). Leave the set based concerns at domain / db level where you can have strong consistency guarantees.

I would strongly advise that message validators, your first line of defense after auth, work exclusively with the message and not have any dependencies on services. (There are always exceptions OC). 

My understanding is that fluent validator objects are quite expensive to instantiate due to compiled expressions etc as such _should_ be singleton wrt to the life time of your app (I actually use static instances). I might be out of date with this understanding, but if I'm correct, the only time you might need instances if there is a some sort setting adjustable at runtime.

mysticmind:
@damianh All your points makes sense and totally agree that fluent validator objects are quite expensive to instantiate. Besides it is best to have clear separation of concerns as you rightly pointed out.

[...]

PR #126, opened 2018-08-02
Update DI registration of validators as scoped GH-123

mysticmind:
Update DI registration of validators as scoped GH-123


mysticmind:
With reference to the thread of comments on #123, closing this.

Issue #145, opened 2019-01-10
Method not allowed when having two routes differing only by verb and parameter name


wastaz:
If I have two routes that differ only by verb and by the name of one or more of the parameters in the route, then it seems like carter will overwrite all routes except the last.

For example, in the module below if you try to call `GET /test/42/blah/d9796528-4631-44e2-b898-6caf2706467d` then you will get `405 method not allowed`. However `PUT /test/42/blah/d9796528-4631-44e2-b898-6caf2706467d` will work with a `200 OK` and the expected response body.

If you rename the `other1` and `other2` parameter in the route, the both the `GET` and `PUT` routes will work as expected.

```

    public class TestModule : CarterModule
    {
        public TestModule()
        {

            Put("/test/{id:int}/blah/{otherid1:guid}", async (request, response, route) =>
            {
                var id = route.As<int>("id");
                var otherid = route.As<Guid>("otherid1");

                response.StatusCode = (int) HttpStatusCode.OK;
                await response.AsJson(new
                {
                    Id = id,
                    Other = otherid,
                    Method = "Put"
                });
            });
            
            Get("/test/{id:int}/blah/{otherid2:guid}", async (request, response, route) =>
            {
                var id = route.As<int>("id");
                var otherid = route.As<Guid>("otherid2");

                response.StatusCode = (int) HttpStatusCode.OK;
                await response.AsJson(new
                {
                    Id = id,
                    Other = otherid,
                    Method = "Get"
                });
            });
        }
    }
```

jchannon:
It must be failing here somehow if that's the case
https://github.com/CarterCommunity/Carter/blob/master/src/CarterExtensions.cs#L73

However I'd expect the tests to fail here for different verbs
https://github.com/CarterCommunity/Carter/blob/master/test/Carter.Tests/TestModule.cs#L74
unless with extra route constraints etc it is an issue.

Probably best to write a test to see if it fails

jchannon:
Hi, have started looking at this however I saw the opposite of what you said, when they both have the same `other1` routes both work. If the GET has `other2` and PUT has `other1` and I issue a GET I get a 405

davidfowl:
cc @JamesNK @rynowak

JamesNK:
Routing doesn't care about the name when matching. How could it? The parameter name is just the name the route value is given when the route has matched.

IRouter routing doesn't have any built in logic for the HTTP method.

jchannon:
[...]

PR #146, opened 2019-01-11
Execute correct handler

jchannon:
Linked to #145


jchannon:
This will be fixed when #149 is merged

PR #154, opened 2019-01-21
Fix naming convetions for AsJson and FromStream

danielmoncada:
Update to fix Async naming in Response extensions. (append Async for methods that return a Task)


JoeStead:
Hi Daniel,

Thanks for the PR, the effort is appreciated!

The decision not to suffix these methods with `Async` was a conscious one in the early days (when it was still known at Botwin). As there are no sync overloads for these methods, it creates unnecessary noise. I know this is a hotly debated opinion, but it is the one Carter took.

Thanks again for the effort, and feel free to contribute in the future, it is appreciated

danielmoncada:
@JoeStead thanks Joe, that makes sense.

Issue #151, opened 2019-01-17
Check for leaking public classes


jchannon:
Things like `SkipEmptyContractResolver` should not be public.

PR #159, opened 2019-02-01
Fixed default json handling

### LINKED ISSUE #155: Disable DefaultValueHandling = DefaultValueHandling.Ignore on the default Json negotiator
By setting `DefaultValueHandling = DefaultValueHandling.Ignore` on the default json, information is easily left off of the response. Consider the following object:

```
public class MyObject 
{
 public string Name {get; set;}
 public int CoolRating {get; set; }
}
```

when it is used:
```
new MyObject {
Name = Jon,
CoolRating = 0
}
```

will result in a json object looking like
```json
{
"name": "Jon",
}
```

where it should actually include the rating field:
```json
{
"name": "Jon",
"CoolRating": 0
}
```

JoeStead:
Fixes #155 

Also a change that affects #151

### DIFF
--- a/src/DefaultJsonResponseNegotiator.cs
+++ b/src/DefaultJsonResponseNegotiator.cs
@@ -20,7 +20,7 @@
             {
                 NamingStrategy = new CamelCaseNamingStrategy()
             };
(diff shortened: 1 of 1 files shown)


jchannon:
Will take a look. This willl effect the OpenApi stuff so will take a look. Probably need to add some tests in the sample.trsts folder to check we get back certain json

JoeStead:
Tests are a bit of a pain, because defaults in C# are defaults, so it will always behave correctly when C# is the client. You'll have to inspect the json in string form

PR #169, opened 2019-02-19
Adds support for user-supplied openapi examples

rcknight:
(As discussed in slack)

Users can supply an example instance on `RouteMetaData` and
`RouteMetaDataResponse`. If supplied, the values on these objects will
be used to populate examples in the OpenApi schemas. If no example
objects are supplied, then this PR still improves example output,
displaying correctly typed values rather than deciding everything is a
string, and correctly handling nested object structures rather than only
including top-level properties in examples.

### DIFF
--- a/src/OpenApi/CarterOpenApi.cs
+++ b/src/OpenApi/CarterOpenApi.cs
@@ -1,8 +1,11 @@
 namespace Carter.OpenApi
 {
     using System;
+    using System.Collections;
     using System.Collections.Generic;
     using System.Linq;
+    using System.Reflection;
+    using System.Text;
     using System.Threading.Tasks;
     using Carter.Request;
     using FluentValidation.Validators;
@@ -196,11 +199,14 @@ namespace Carter.OpenApi
                         Type responseType;
                         var responseTypeName = string.Empty;
                         var singularTypeName = string.Empty;
+                        object singleExample = null;
 
                         if (valueStatusCode.Response.IsArray() || valueStatusCode.Response.IsCollection() || valueStatusCode.Response.IsEnumerable())
                         {
                             arrayType = true;
                             responseType = valueStatusCode.Response.GetElementType();
+                            singleExample = ((IEnumerable<object>)valueStatusCode.ResponseExample)?.FirstOrDefault();
+
                             if (responseType == null)
                             {
                                 responseType = valueStatusCode.Response.GetGenericArguments().First();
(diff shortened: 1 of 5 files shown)


rcknight:
The build will be failing because I didn't yet update the expected test output. Even in this default case where no example instance is supplied, the example object has improved:

new:
``` json
"example": {
	"name": "Example",
    "id": 0,
    "age": 0
}
```

old:
``` json
"example": {
    "name": "",
    "id": "",
    "age": ""
}
```

src/OpenApi/CarterOpenApi.cs
```diff
@@ -170,11 +173,14 @@ private static void CreateOpenApiResponseBody(OpenApiDocument document, KeyValue
                         Type responseType;
                         var responseTypeName = string.Empty;
                         var singularTypeName = string.Empty;
+                        object singleExample = null;
 
                         if (valueStatusCode.Response.IsArray() || valueStatusCode.Response.IsCollection() || valueStatusCode.Response.IsEnumerable())
                         {
                             arrayType = true;
                             responseType = valueStatusCode.Response.GetElementType();
+                            singleExample = ((IEnumerable<object>)valueStatusCode.ResponseExample)?.FirstOrDefault();
```
jchannon: Shouldn't this be set if the response is not an array?

rcknight: Hmm you're right it's weird. I think I put this here because on L201 the existing code constructs a schema for a single item even in the array case, but looking now that never actually gets used in  the case of an array. I think some rearranging to move that inside the if at the bottom can make this go away.

rcknight: Fixed in latest commit

src/OpenApi/CarterOpenApi.cs
```diff
@@ -226,24 +200,38 @@ private static void CreateOpenApiResponseBody(OpenApiDocument document, KeyValue
...
+
                                 document.Components.Schemas.Add(responseTypeName, schema);
                             }
                             else
                             {
                                 document.Components.Schemas.Add(responseTypeName,
-                                    new OpenApiSchema { Type = "array", Items = new OpenApiSchema { Reference = new OpenApiReference { Id = singularTypeName, Type = ReferenceType.Schema } } });
+                                    new OpenApiSchema {
+                                            Type = "array",
+                                            Example = example,
```
jchannon: This needs to be an array of `example` or we decide we don't do examples for arrays of items as the arrays have a link to the singular item which has an example, see here https://github.com/CarterCommunity/Carter/blob/master/test/Carter.Samples.Tests/OpenAPITests.Should_return_Carter_approved_OpenAPI_json.approved.txt#L156 Thoughts?

jchannon: Any thoughts?

test/Carter.Tests/OpenApiTests.cs
```diff
@@ -0,0 +1,39 @@
...
+                }
+            );
+
+            using (var body = new MemoryStream())
+            {
+                var ctx = new DefaultHttpContext();
+                ctx.Response.Body = body;
+                await openApiDelegate.Invoke(ctx);
+                var responseBody = Encoding.UTF8.GetString(body.ToArray());
+                Assert.Contains("Docs", responseBody);
```
munies4530: 76561198006963719

jchannon:
Recent OpenAPI changes from @williamhbell don't create the "Example" json objects as swagerui does it for you

PR #190, opened 2019-03-22
Include optional argument to return null fields in json response

cbuddeke:
I have a need to be able to return fields even if they are null, so I included an optional argument to allow that.


jchannon:
Thanks for this.

I think for now you'll have to register your own response negoatiator to
get this feature.

What I'd like to do in the future is expose a jsonsettings feature in the
carter options.

However, in asp.net core 3 they are changing json serialization and
removing json.net dependency so wlll be interesting to see what they do and
i could maybe bubble it up to carter possibly.

Issue #194, opened 2019-06-27
Invalid cast from 'System.String' to 'System.Guid'.


cocojiambo:
Throws "Invalid cast from 'System.String' to 'System.Guid'." error when binding a System.Guid from the Request.Body on PUT requet.

See https://github.com/CarterCommunity/Carter/pull/107#issuecomment-506275495 for more info

cocojiambo:
Issue fixed in commit 89f3161

PR #196, opened 2019-06-27
Issue #194 Invalid cast from 'System.String' to 'System.Guid'

cocojiambo:
Fix "Invalid cast from 'System.String' to 'System.Guid'." issue. Happens when submitting a PUT request that has Guids as part of its body.

### DIFF
--- a/src/ModelBinding/BindExtensions.cs
+++ b/src/ModelBinding/BindExtensions.cs
@@ -62,6 +62,17 @@ namespace Carter.ModelBinding
                         }
                         return val.Value.Select(y => Convert.ChangeType(y, colType));
                     }
+
+                    // Convert.ChangeType (below) chokes on Guids, so doing it here instead.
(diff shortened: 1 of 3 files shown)


cocojiambo:
@jchannon how long does it usually take for the new package to be available? Or have you got a pipeline for it?

jchannon:
Thanks for your contribution! 🎉 👍 

Releases vary however you can use a built package on Carter's nuget feed. If you add this nuget.config file next to your solution or amend your global nuget.config you'll be able to pick up a package with your fix in it.

```
<?xml version="1.0" encoding="utf-8"?>
<configuration>
    <packageSources>
        <add key="carter" value="https://ci.appveyor.com/nuget/carterci/" />
    </packageSources>
```

You'll see that version 4.2.0-ci108 is available that has your changes in it and anything else since the 4.2.0 release

jchannon:
Changes since the last release can be seen here https://github.com/CarterCommunity/Carter/milestone/16?closed=1

PR #199, opened 2019-07-14
Fix exception with form binding and nullable types

### LINKED ISSUE #198: Invalid cast from 'System.String' to 'System.Nullable`1[[System.Boolean, System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]]'.
Fails to bind a string to bool. Blows up with the error:

Invalid cast from 'System.String' to 'System.Nullable`1[[System.Boolean, System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]]'.

image

welshdave:
Fixes #198

### DIFF
--- a/src/ModelBinding/BindExtensions.cs
+++ b/src/ModelBinding/BindExtensions.cs
@@ -2,6 +2,7 @@ namespace Carter.ModelBinding
 {
     using System;
     using System.Collections.Generic;
+    using System.ComponentModel;
     using System.IO;
     using System.Linq;
     using System.Threading.Tasks;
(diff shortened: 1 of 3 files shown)


jchannon:
Looking good. Will take a proper look tomorrow. Thanks!

jchannon:
This is up to you but I see a few comments here about some TypeConverters not working in certain circumstances and would like to expand the tests a bit to ensure that we can convert those. If you want to take a look let me know or I'll merge then take a look and fix any issues if they exist https://www.hanselman.com/blog/TypeConvertersTheresNotEnoughTypeDescripterGetConverterInTheWorld.aspx

welshdave:
Wouldn't be able to look for a few days (travelling in to office to prove I still exist), but could look Thursday or Friday?

jchannon:
Cool!

welshdave:
So TypeConverters have a few problems - don't think it's worth using them. I've updated and added a few more tests.

jchannon:
What was the issues with them?

welshdave:
They lose precision when converting datetime (loses milliseconds), seem to be a lot more picky about date formatting than other conversion methods. 

And for number formatting, strings like "1,234.00" don't convert (try converting that to a decimal, it fails with the TypeConverter approach, works fine with Convert.ChangeType).

jchannon:
Is it worth adding some tests for decimals?

welshdave:
Yeah, I'll add some now

jchannon:
Thanks @welshdave Awesome stuff!

PR #205, opened 2019-09-16
Correcting the code to handle ulong and nullable simple types

williamhbell:
As discussed today, this pull request is to address issues that affect ulong and ulong? types.  Other numeric types have been added for completeness.


jchannon:
Thanks 👍

PR #210, opened 2019-10-26
Benchmarks

jchannon:


jchannon:
Wonder if we need to create a load of modules to get better understanding on the metrics. Thoughts?

ritasker:
The module I register in the test is the hello world one.
I think if there is a module with multiple endpoints using the different HTTP methods it will be more realistic.

jchannon:
But is profiling startup on 6 routes for example enough?

JoeStead:
Have lots of different startups?

ritasker:
Could use `WithModule` on the `CarterConfigurator` to register different modules in seperate tests and see the difference.

jchannon:
I don't want to test different modules in this scenario. I just want to see
the benchmarks of calling Add/Use Carter then call it again on the
aspnetcore3 stuff to see if it has improved/degraded

ritasker:
I have removed the stub module and tried to benchmark the Add/UseCarter methods.
The UseCarter method is being a bit tricky, so the benchmark is failing ATM.

ritasker:
`<PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0-beta-63127-02" PrivateAssets="All" />` I have to remove this locally its also breaking the build.

benchmarks/Carter.Benchmarks/Program.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace Carter.Benchmarks
+{
+    using BenchmarkDotNet.Running;
+
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            BenchmarkSwitcher.FromAssembly(typeof(Program).Assembly).Run(args);
```
adamsitnik: Hi @ritasker 

I saw your question at https://gitter.im/dotnet/BenchmarkDotNet

If you are getting timeout on a CI then one thing you could consider would be to change default BenchmarkDotNet settings.

BDN does not assume anything about the code (herutistics/caching strategies) and it runs at least 6 warmup iterations. You could set it to 1.
Moreover, the default IterationTime is 500ms, you could set it to 250ms.
Another important thing is that BDN runs the benchmark until the results become stable or it reaches the default max iteration count (100). You could set it to 20.

```cs
BenchmarkSwitcher.FromAssembly(typeof(Program).Assembly).Run(args, CreateConfig());

private static IConfig CreateConfig()
{
    var job = Job.Default
        .WithWarmupCount(1)
        .WithIterationTime(TimeInterval.FromMilliseconds(250))
        .WithMaxIterationCount(20);

    return DefaultConfig.Instance
        .With(job.AsDefault()); // tell BDN that this are the default settings (so things passed as console line arguments are applied to this job)
}
```

sebastienros:
What about adding TechEmpower benchmarks instead?

jchannon:
Techempower already has them although they need updating with the 5.0.0 release

jchannon:
https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/frameworks/CSharp/carter

Needs some love there with latest C#/SDK etc

jchannon:
What shall we do here?

PR #213, opened 2019-11-04
OpenApi requested changes

williamhbell:
This pull request now includes a fix for arrays and simple nullable types.  Multipart-forms are also handled.  The Carter tests pass and the code has also been tested with a live application, where the OpenApi out is used to generate Angular that is used to interact with the microservices.  If this code is acceptable, code to allow a BindAndValidate action with a mulitpart form post can be provided for discussion.


samples/CarterSample/Features/Actors/FileForm.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Carter.OpenApi;
+
+namespace CarterSample.Features.Actors
+{
+    public class FileForm
+    {
+        public File File { get; set; }
+
+        [ApiSchemaAttributes(Format = "binary")]
```
jchannon: Does this need to be an attribute? Could this property not be deleted and then in PostFileForm it sets the Format property on RouteMetaDataRequest and then when /openapi is being called it uses the RouteMetaDataRequest.Format property to assign the Format property of the OpenApiSchema?

williamhbell: The format type needs to belong to the instance of the schema, rather than the schema itself.  This implies that it is tricky to use the RouteMetaDataRequest to hold it.  In the example above, there are many instances of the type "System.String", but only one is the property of File and has binary format.

test/Carter.Samples.Tests/OpenAPITests.Should_return_Carter_approved_OpenAPI_json.approved.txt
```diff
@@ -229,78 +235,62 @@
...
-        }
-      },
-      "Actors": {
-        "type": "array",
-        "items": {
-          "$ref": "#/components/schemas/Actor"
-        }
-      },
       "Actor": {
-        "required": [
```
jchannon: We shouldn't lose the validation fields like required and minLength and excluseiveMinimum

test/Carter.Samples.Tests/OpenAPITests.Should_return_Carter_approved_OpenAPI_json.approved.txt
```diff
@@ -229,78 +235,62 @@
...
-          "name"
-        ],
         "type": "object",
         "properties": {
           "name": {
-            "minLength": 1,
```
jchannon: We shouldn't lose the validation fields like required and minLength and excluseiveMinimum

test/Carter.Samples.Tests/OpenAPITests.Should_return_Carter_approved_OpenAPI_json.approved.txt
```diff
@@ -229,78 +235,62 @@
...
             "type": "string"
           },
           "id": {
             "type": "integer"
           },
           "age": {
-            "minimum": 0,
-            "exclusiveMinimum": true,
```
jchannon: We shouldn't lose the validation fields like required and minLength and excluseiveMinimum

test/Carter.Samples.Tests/OpenAPITests.Should_return_Carter_approved_OpenAPI_json.approved.txt
jchannon: Having debugged this , for some reason Properties is empty which is looped over in `AddValidationInformation`

jchannon:
Merged via https://github.com/CarterCommunity/Carter/commit/a25afd8fe3f47e854dadf68b3bf1f533e61e0e1d

The git history was a bit wonky

Thanks for all your help here. Brilliant work, thanks!! 🎉 🎉 👍

PR #216, opened 2019-11-25
Options...

jchannon:
Pinging @davidfowl


jchannon:
Pinging @alefranz

samples/CarterSample/Startup.cs
```diff
@@ -17,7 +13,11 @@ public void ConfigureServices(IServiceCollection services)
         {
             services.AddSingleton<IActorProvider, ActorProvider>();
 
-            services.AddCarter();
+            services.AddCarter(options => options.OpenApi = new OpenApiOptions("Carter <3 OpenApi", new[] { "http://localhost:5000" }, new Dictionary<string, OpenApiSecurity>
```
davidfowl: Don't make this an expression. It looks noisy and the line is too long.

jchannon: Pass an instance in instead?

davidfowl: I think so.

davidfowl: Wait, no, I meant the OpenAPI part.

jchannon: done on a new commit!

src/Carter/CarterOptions.cs
```diff
@@ -88,8 +82,11 @@ public enum OpenApiIn
     public enum OpenApiSecurityType
     {
         apiKey,
```
davidfowl: Casing?

samples/CarterSample/Startup.cs
```diff
@@ -26,37 +32,16 @@ public void Configure(IApplicationBuilder app, IConfiguration config)
             config.Bind(appconfig);
```
davidfowl: Configuration isn't being used here.

jchannon: Rejigged Configuration in the sample to use the new options approach in newer commit

Issue #219, opened 2019-12-02
OpenApi Scheme mapping for Requests


Jaxelr:
Hi Carter Community,

Newbie here, so pardon any ignorance in advance. I am having trouble generating a consistent json to consume a Post method, using the generated OpenApi module.

I managed to recreate the issue using the Post method AddActor on the ActorsModule inside Carter's samples (https://github.com/CarterCommunity/Carter/blob/master/samples/CarterSample/Features/Actors/ActorsModule.cs#L43-L58). 

https://github.com/CarterCommunity/Carter/blob/58b8109929903f9420bc4db1aeb4497b0e6932ed/samples/CarterSample/Features/Actors/ActorsModule.cs#L43-L58

![image](https://user-images.githubusercontent.com/5913008/69994577-87d34f80-1524-11ea-8e98-60cc735ad100.png)

Using the request generated from the openapi path, the result is a validation error from fluent, since the casings are different:

![image](https://user-images.githubusercontent.com/5913008/69994628-a1749700-1524-11ea-83a0-85c604cac121.png)

Ive attached the OpenApi json generated:

``` json
{
  "openapi": "3.0.1",
  "info": {
    "title": "Carter <3 OpenApi",
    "version": "3.0.0"
  },
  "servers": [
    {
      "url": "http://localhost:5000"
    }
  ],
  "paths": {
    "/actors": {
      "post": {
        "tags": [
          "Actors"
        ],
        "description": "Create an actor in the system",
        "operationId": "Actors_AddActor",
        "requestBody": {
          "content": {
            "application/json": {
              "schema": {
                "$ref": "#/components/schemas/Actor"
              }
            }
          }
        },
        "responses": {
          "201": {
            "description": "Created Actors"
          }
        }
      }
    }
  },
  "components": {
    "schemas": {
      "Actor": {
        "required": [
          "name"
        ],
        "type": "object",
        "properties": {
          "name": {
            "minLength": 1,
            "type": "string"
          },
          "id": {
            "type": "integer"
          },
          "age": {
            "minimum": 0,
            "exclusiveMinimum": true,
            "type": "integer"
          }
        }
      }
    }
}
```

Given the mismatch is generated using the CamelCase static method inside the CarterOpenApi class:

https://github.com/CarterCommunity/Carter/blob/58b8109929903f9420bc4db1aeb4497b0e6932ed/src/Carter/OpenApi/CarterOpenApi.cs#L144

and:

[...]

PR #220, opened 2019-12-03
Added Property Insensitive Casing to request deserialization (#219)

Jaxelr:
This is done in order to avoid binding failures on scenarios where the
casing is different on the request.


jchannon:
Thanks @Jaxelr 👍 ❤️

PR #225, opened 2019-12-09
Alternate binder

gkinsman:
This PR adds a new `IModelBinder` interface that allows substitution of an alternative model binding mechanism. 

The default binder uses `System.Text.Json`, but has some shortcomings such as that it cannot bind properties with private setters. 

This PR also implements a model binder that uses `Newtonsoft.Json` for backwards compatibility. It doesn't come out of the box, but is in the tests.


src/Carter/CarterExtensions.cs
```diff
@@ -191,6 +192,8 @@ private static void WireupCarter(this IServiceCollection services, DependencyCon
             }
 
             services.AddSingleton<IResponseNegotiator, DefaultJsonResponseNegotiator>();
+
+            services.AddSingleton<IModelBinder, DefaultJsonModelBinder>();
```
jchannon: Can you add a method GetModelBinders and do what GetResponseNegotiators does and finds binders and ignores the default one, that way people can just drop theres in

test/Carter.Tests/Modelbinding/NewtonsoftBinding/NewtonsoftTests.cs
```diff
@@ -0,0 +1,59 @@
...
+                ContractResolver = new NewtonsoftJsonUtils.PrivateSetterCamelCasePropertyNamesContractResolver(),
+                Converters = new JsonConverter[] {new StringEnumConverter()},
+            };
+            
+            var server = new TestServer(
+                new WebHostBuilder()
+                    .ConfigureServices(x =>
+                    {
+                        x.AddCarter(configurator: c =>
+                            c.WithModule<NewtonsoftModule>());
```
jchannon: I wonder if we expose `WithBinder` option and provide a JSON.NET Binder too so people can use that when settings up Carter in their apps

jchannon:
Thanks! 👍

PR #230, opened 2019-12-28
Introduced TryValidate<T>() extension method

klym1:
...for `HttpRequest` which accompanies `Validate<T>()`, but is more flexible, because it allows to explicitly handle the case when the model's validator is missing.

### DIFF
--- a/src/Carter/ModelBinding/ValidationExtensions.cs
+++ b/src/Carter/ModelBinding/ValidationExtensions.cs
@@ -26,6 +26,29 @@ namespace Carter.ModelBinding
         }
 
         /// <summary>
+        /// Performs validation on the specified <paramref name="model"/> instance
+        /// </summary>
+        /// <typeparam name="T">The type of the <paramref name="model"/> that is being validated</typeparam>
+        /// <param name="request">Current <see cref="HttpRequest"/></param>
+        /// <param name="model">The model instance that is being validated</param>
+        /// <param name="result"><see cref="ValidationResult"/></param>
+        /// <returns>true if there's a validator associated with the <paramref name="model"/></returns>
+        public static bool TryValidate<T>(this HttpRequest request, T model, out ValidationResult result)
+        {
+            var validatorLocator = request.HttpContext.RequestServices.GetService<IValidatorLocator>();
+            var validator = validatorLocator?.GetValidator<T>();
+
+            if (validator == null)
+            {
+                result = null;
+                return false;
+            }
+
+            result = validator.Validate(model);
+            return true;
+        }
+
+        /// <summary>
         /// Retrieve formatted validation errors
         /// </summary>
         /// <param name="result"><see cref="ValidationResult"/></param>


jchannon:
Thanks!

I’m not against this but it has made me think. why would you want a Boolean
when you already have it in the Validate method via the returned object
IsValid property?

I wonder if Carter should throw if no validator is found instead?

klym1:
> why would you want a Boolean when you already have it in the Validate method via the returned object IsValid property

The answer is - separation of concerns, aka SRP. There are actually 2 quite distinct cases - when a validator exists for a model and when it does not. 
In the first case, we can tell what the validation result is because we can take the validator and check the model against it.
In the second case, we can't get a validation result at all (there's no validator to check against). 
Therefore it would be wrong to claim `IsValid=false` because there's no result.

That's why these 2 paths must be handled differently.

> I wonder if Carter should throw if no validator is found instead?

Back in Nancy times - a model with a missing validator would be counted as a valid model, so I think better to keep such behavior in Carter. Also, with *Try* semantics we can leave that decision up to end-users. 

For example, users might choose not to throw an exception, but rather return an error (which is preferable in most cases):

```csharp
//if no validator exists - skip the section, count a model as valid
if (req.TryValidate(model, out var validationResult) && !validationResult.IsValid)
{
    // validator exists, the model is not valid, return the error
}
```

jchannon:
At the moment Carter doesn’t return a valid result on no validator found.
Can’t remember how Nancy did it. Would need to look.

Think Carter should throw actually and we can have the TryValidate that
returns false.

Would need to alter and add some tests for this PR

klym1:
> At the moment Carter doesn’t return a valid result on no validator found
And that's a bad approach. Let me explain why.

Consider two scenarios:
1) typical scenario: Nancy project migrating to Carter. It uses Fluent validator, but not all the models have validators (that works perfectly with Nancy). In order to make the project running with Carter developers would have either to create at least `AbstractValidator<>` stubs for all the models without validators or write own code which would pass the validation when there's no validator.

2) A new project built on top of Carter. It's been under active development and then developers decide to use FluentValidator. But they're facing the same dilemma: write validators for all the models at once or write their own code that would wrap the validation logic.

jchannon:
But why call Validate when you have no validator to validate the model

klym1:
Validating a model is a very boilerplate-ish operation, i.e. it requires a very similar code to be written in every handler, so very often it is moved to the infrastructure level, where a framework would call it for every handler. And I'm not sure if it's possible to find out whether there's a validator for a model by using e.g. FluentValidator capabilities (even if it is, I'd still prefer to have some consistent mechanism for that, that would work consistently across other validation libs).

jchannon:
Thanks!

jchannon:
Just thought this through some more and if you don't want validate to return a failure don't call it. If you have copy/pasted code that has BindAndValidate don't call it, just call Bind

jchannon:
Have reverted the code.

PR #231, opened 2019-12-28
Tweak MinVer

khalidabuhakmeh:
Marks MinVer as a  which makes sure it doesn't show up in
the final build as a dependency. Also, only runs MinVer on
builds.


jchannon:
Thanks ❣👍

Issue #229, opened 2019-12-23
Add option to disable OpenApi feature which is enabled by default


alifxyd:

jchannon:
Is it causing issues? I’m not against adding the option though. Feel free
to add the property to the option class and return 404 if disabled in a PR?
:)

alifxyd:
I don't want to expose the `/openapi` endpoint when there is no need for it. And it would be nice to have it configurable. As a workaround I can create a middleware to reject all the requests to it, but it's better to not map it on startup.

Okay, will try to find some time for this a bit later.

Thanks

PR #232, opened 2019-12-28
Add Enabled To OpenAPI Options

khalidabuhakmeh:
Allows you to disable the GET endpoint for OpenAPI
for thos folks who don't want it as part of their
Carter instance.

 #229

Should be backward compatible safe.


jchannon:
You’re on fire! 🔥 

Thanks

PR #234, opened 2020-01-13
Accept route metadata instance

heerup:
I changed to referring an interface `IRouteMetadata` instead of the abstract class (the abstract base class is still there) to allow for more composition scenarios.
Then I Added the option to pass an instance of `IRouteMetadata` instead of using the type parameter.


jchannon:
Could you add an example to the sample app to show how you see having this
interface put in providing value or maybe document it in the PR, whatever
you think is easier for you :)

PR #249, opened 2020-06-04
Corrections to address openapi generation failure for nullable types

williamhbell:
These fixes are needed to address a bug that affects some objects that contain simple nullable types, such as "int?".


jchannon:
Thanks! 👍

PR #251, opened 2020-06-17
Change module scoped logger

gkinsman:
This PR changes the logger used in the request lambda in `CarterExtensions` to use a category of `Carter` instead of the full name of the module type. The module type name is then added as a scope in the request handler.

This makes it possible to filter out the `Executing module route handler for` message filtering out `Carter` category instead of having to do it on each module type. It should generally be possible to filter out carter logs with the `Carter` category I think.

I've addded `MartinCostello.Logging.XUnit` which connects XUnit's `ITestOutputHelper` with ANC logging so the logs will show up in the test output, so it's easy to see what is being output. Currently only added it to the `CarterModuleTests`, not sure if we want/need this behaviour elsewhere. I think its useful.

### DIFF
--- a/src/Carter/CarterExtensions.cs
+++ b/src/Carter/CarterExtensions.cs
@@ -83,11 +83,6 @@ namespace Carter
         {
             return async ctx =>
             {
-                using var moduleScope = logger.BeginScope(new Dictionary<string, object>()
-                {
-                    ["ModuleType"] = moduleType.Name
-                });
-                
                 // Now in per-request scope
                 var module = ctx.RequestServices.GetRequiredService(moduleType) as CarterModule;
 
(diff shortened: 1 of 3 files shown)


src/Carter/CarterExtensions.cs
```diff
@@ -86,6 +83,11 @@ public static IEndpointConventionBuilder MapCarter(this IEndpointRouteBuilder bu
         {
             return async ctx =>
             {
+                using var moduleLogger = logger.BeginScope(new
```
jchannon, line 86: What does this do?

test/Carter.Tests/Carter.Tests.csproj
```diff
@@ -8,4 +8,7 @@
   <ItemGroup>
     <ProjectReference Include="..\..\src\Carter\Carter.csproj" />
   </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="MartinCostello.Logging.XUnit" Version="0.1.0" />
```
jchannon, line 12: I think if we add this we need to have a test to check logs but then i'm not sure we want to have unit tests for logs?

PR #256, opened 2020-08-02
Make API version configurable

### LINKED ISSUE #255: API Version hardcoded in SwaggerUI to 3.0.0
Is there a a reason why the API version is hard coded to 3.0.0 ?

The line I am referring is this: https://github.com/CarterCommunity/Carter/blob/33e22c6d3f7f5daa0aa63673749ff79d70845993/src/Carter/OpenApi/CarterOpenApi.cs#L33

It is my understanding that this version is the API version not the OpenAPI Spec Version.

JoeStead:
Fixes #255 by making the API version configurable

### DIFF
--- a/src/Carter/CarterOptions.cs
+++ b/src/Carter/CarterOptions.cs
@@ -33,6 +33,11 @@ namespace Carter
         public string DocumentTitle { get; set; } = "Carter <3 OpenApi";
 
         /// <summary>
+        /// The version of the API defined by the OpenApi document
+        /// </summary>
+        public string Version { get; set; } = "1.0.0";
+
+        /// <summary>
         /// The servers property of the OpenApi document
         /// </summary>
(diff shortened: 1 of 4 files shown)


JoeStead:
I have no clue WTF is going on with that diff tool stuff. Gonna go right ahead and say that test needs changing

jchannon:
Can you add a `openapi` property which should be 3.0.0. The test is running a comparison of a hard coded json doc to what calling a route that will return a openapi doc

JoeStead:
No, because it isn't available on the type

jchannon:
Looks like it might just add it
https://github.com/microsoft/OpenAPI.NET/blob/vnext/src/Microsoft.OpenApi/Models/OpenApiDocument.cs#L75

jchannon:
Cheers! 👍

PR #258, opened 2020-08-30
Provide a message when Validate cannot resolve a Validator

ggmueller:
At the moment Carter throws an `NullReferenceException` when no validator can be resolved from the ValidatorLocator. The `NullReferenceException` contains no further information as to why this Exception was thrown.

I suggest to add an informative message, so that the developer can find the reason for the exception. 

I also suggest throwing an `InvalidOperationException` here, as null is not dereferenced in this case.


jchannon (approved):

jchannon:
Thanks!! 👍

Issue #262, opened 2020-10-24
OpenAPI Enum Support


Dreamwalker666:
OpenAPI supports enums of string values

Currently enums output as  `{}` in `components.schema`.  I think these should output using string and the enum values using the enum specification

https://swagger.io/docs/specification/data-models/enums/

PR #264, opened 2020-10-27
Solution to support Enum's in OpenAPI for Issue #262

Dreamwalker666:
Created a possible solution for issue #262

### DIFF
--- a/.github/workflows/release-notes.yml
+++ b/.github/workflows/release-notes.yml
@@ -18,6 +18,6 @@ jobs:
         cp  temp_release_notes/*.md releasenotes
         git config user.name Jonathan Channon
         git config user.email jonathan.channon@gmail.com
-        git add releasenotes/
+        git add .
         git commit -m "Generated Release Notes"
         git push 
--- a/samples/CarterSample/Features/Actors/Actor.cs
+++ b/samples/CarterSample/Features/Actors/Actor.cs
@@ -5,5 +5,14 @@ namespace CarterSample.Features.Actors
         public string Name { get; set; }
         public int Id { get; set; }
         public int Age { get; set; }
+        
+        public ActorStatus Status { get; set; }
+    }
+    
+    public enum ActorStatus
+    {
+        Active,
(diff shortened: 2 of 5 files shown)


JoeStead:
I wonder if there should be some config option for this. Enums in C# are a bit weird imo, it makes sense (mostly) for them to be serialised as strings, but I have come across solutions using integers over the wire (this is a bad idea, but we should cater for it)

JoeStead (approved):
That said, the implementation of this seems pretty sound

Dreamwalker666:
> I wonder if there should be some config option for this. Enums in C# are a bit weird imo, it makes sense (mostly) for them to be serialised as strings, but I have come across solutions using integers over the wire (this is a bad idea, but we should cater for it)

I honestly have the same thoughts but decided to first implement in the way OpenAPI expects.  I really think its a bad idea to use numbers! (pesky magic numbers)

Maybe use a custom attribute to override the behavior?  I think strings should be the default (we should encourage good API design) but allow it to expose the numeric values.  I guess as well turn off (but currently it gives out a blank object which isn't useful at all)

JoeStead:
Ooof, would be wary of magic attributes. I think string defaults make sense, perhaps an extension to the CarterOptions to allow an override? I thinking turning it off makes no sense, so either string (default) or number (override)

Dreamwalker666:
> Ooof, would be wary of magic attributes. I think string defaults make sense, perhaps an extension to the CarterOptions to allow an override? I thinking turning it off makes no sense, so either string (default) or number (override)

I only said attribute as you may have some enums that require number but not all.  Also the number they will still have to be a string in the schema as enum only supports string from what I've read.

jchannon (approved):

jchannon:
Can you do a rebase please before I merge!?

Dreamwalker666:
@jchannon ok sorted it was expecting that issue due to the other branch :)

jchannon:
Nice! Thanks 👍

Issue #263, opened 2020-10-24
OpenApi JsonIgnore Support


Dreamwalker666:
I think the `JsonIgnore` attribute should be honored so members marked with this aren't output in the schema.

Now of course this brings in the which serialize scenario!  So I propose its stored as an option something like:

`Type IgnoreAttribute { get; set; } = typeof(System.Text.Json.Serialization.JsonIgnoreAttribute);`

(And of course defaulting to .NET core inbuilt)

Now this does mean original behavior can be achieved by setting the setting to `null` as well.  And also means Json.NET can be supported by swapping the value.

Another options is as standard its set to `null` and you have to opt into members being ignored

PR #265, opened 2020-10-27
Solution to use JsonIgnore on OpenAPI schema objects #263a

Dreamwalker666:


Dreamwalker666:
Opps not sure how that a got on the end of the issue number.  This obviously should like to #263 :)

src/Carter/OpenApi/CarterOpenApi.cs
```diff
@@ -368,9 +371,19 @@ private static void AddSchema(Dictionary<string, SchemaElement> navigation, Open
                 
                 foreach (var memberKeyValue in keyValuePair.Value.DataMembers)
                 {
-                    var propertySchema = SchemaFromElement(memberKeyValue.Value);
                     var propertyInfo = keyValuePair.Value.ElementType.GetProperties().Where(o => CamelCase(o.Name) == memberKeyValue.Key).SingleOrDefault();
                     object[] attribute = propertyInfo.GetCustomAttributes(typeof(ApiSchemaAttributes), true);
+
+                    if (options.SchemaIgnoreAttribute != null)
```
jchannon, line 377: Won't this always be not null?

src/Carter/CarterOptions.cs
```diff
@@ -56,6 +56,13 @@ public class OpenApiOptions
         /// Determines if OpenAPI should be mapped and enabled. True by default.
         /// </summary>
         public bool Enabled { get; set; } = true;
+        
+        /// <summary>
+        /// Attribute used to ignore peoperties from the OpenAPI schema
```
jchannon, line 61: Typo 😁

src/Carter/OpenApi/CarterOpenApi.cs
```diff
@@ -351,7 +351,10 @@ private static OpenApiDocument CreateDocument(CarterOptions options)
         /// </summary>
         /// <param name="navigation">The input SchemaElements that have been read from the input classes.</param>
         /// <param name="document">The OpenApiDocument that is to be filled with the schema information.</param>
-        private static void AddSchema(Dictionary<string, SchemaElement> navigation, OpenApiDocument document, HttpContext context)
+        /// <param name="context"></param>
```
jchannon, line 354: Can we fill these parameters info in?

jchannon (changes requested):
Looks good! Thanks

src/Carter/OpenApi/CarterOpenApi.cs
Dreamwalker666: You can set the option to null to turn off the feature.  I did it like that for backwards compatibility.  But thinking on it now who would use the attribute and expect it not to get ignored?

samples/CarterSample/Features/Actors/Actor.cs
```diff
@@ -1,9 +1,14 @@
 namespace CarterSample.Features.Actors
 {
+    using System.Text.Json.Serialization;
```
JoeStead, line 3: Should this be a carter-specific attribute instead? Given Carter can be configured with it's JSON Serialiser, it's possible to mix-and-match with it here. If using Newtonsoft this seems a bit odd to me

src/Carter/CarterOptions.cs
```diff
@@ -56,6 +56,13 @@ public class OpenApiOptions
         /// Determines if OpenAPI should be mapped and enabled. True by default.
         /// </summary>
         public bool Enabled { get; set; } = true;
+        
+        /// <summary>
+        /// Attribute used to ignore properties from the OpenAPI schema
+        ///
+        /// Note: Set this to null to disable.
+        /// </summary>
+        public Type SchemaIgnoreAttribute { get; set; } = typeof(System.Text.Json.Serialization.JsonIgnoreAttribute);
```
JoeStead, line 65: Ahh, I see it's possible to override. Ignore the last comment :)

Dreamwalker666: Yeah didn't want to tie it to just the internal .net attribute.  Too many serialisers out there :)

jchannon:
Thanks! 👍

PR #266, opened 2020-10-27
Test failures

NiklasHansen:
4 tests were failing due to assuming english formatting of dates and decimals. When run on systems with non-english regional settings, these tests would fail. I have corrected the tests to cater for different regional settings.


jchannon:
Thanks! 👍

PR #272, opened 2021-01-14
Fixed some copy-and-paste errors in comments

Earl-Brown:
The summary for some non-POST methods said they were for POST request.  I updated them to match the method being registered

### DIFF
--- a/src/Carter/CarterModule.cs
+++ b/src/Carter/CarterModule.cs
@@ -244,7 +244,7 @@ namespace Carter
         }
 
         /// <summary>
-        /// Declares a route for POST requests
+        /// Declares a route for PUT requests
         /// </summary>
         /// <param name="path">The path for your route</param>
         /// <param name="handler">The handler that is invoked when the route is hit</param>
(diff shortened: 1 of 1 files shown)


src/Carter/CarterModule.cs
```diff
@@ -256,7 +256,7 @@ protected IEndpointConventionBuilder Put(string path, RequestDelegate handler)
         }
 
         /// <summary>
-        /// Declares a route for POST requests
+        /// Declares a route for PUT requests
```
jchannon, line 259: Head?

src/Carter/CarterModule.cs
```diff
@@ -311,7 +311,7 @@ protected IEndpointConventionBuilder Head(string path, RequestDelegate handler)
         }
 
         /// <summary>
-        /// Declares a route for POST requests
+        /// Declares a route for HEAD requests
```
jchannon, line 314: Patch?

jchannon (changes requested):
Thanks!!

jchannon:
CarterModule will no longer exist for next release. 

Thanks for the PR though 👍

PR #277, opened 2021-04-28
Bind querystring to array

### LINKED ISSUE #276: Could not bind an array of query param values with Request.Query.As<int[]>("Id")
```csharp
public sealed class AppCarterModule : CarterModule
{
    private static readonly Actor[] Actors = new[]
    {
        new Actor() { Id = 1, Name = "First"},
        new Actor() { Id = 2, Name = "Second"},
        new Actor() { Id = 3, Name = "Third"},
    };
    public AppCarterModule()
    {
        Get("/actors", async ctx =>
        {
            IEnumerable<Actor> result = Actors;
            // ids is always null for "/actors?Id=1&Id=3"
            var ids = ctx.Request.Query.As<int[]>("Id");
            if (ids != null)
            {
                result = result.Where(x => ids.Contains(x.Id));
            }
            await ctx.Response.Negotiate(result);
        });
    }
}
```

`GET /actors?id=1&id=3`
returns all 3 records instead of 2 because `ctx.Request.Query.As<int[]>("Id")` returns `null` instead of `[1,3]`

Repository with a failing test
https://github.com/deilan-issues-samples/Carter.Issues

```csharp
public async Task Test1()
        {
            var client = _factory.CreateClient();
            var result = await client.GetAsync("/actors?Id=1&Id=3");
            var str = await result.Content.ReadAsStringAsync();
            var array = JsonConvert.DeserializeObject<dynamic[]>(str);

            Assert.Equal(2, array.Length);
        }
```

jchannon:
Fix #276﻿


jchannon:
No need for PR - can use `var id = ctx.Request.Query.AsMultiple<int>("id");`

PR #282, opened 2021-11-12
.NET 6 Baby!

jchannon:
This is a large breaking change for .NET6.

`CarterModule` has been replaced with `ICarterModule`
`OpenAPI` support has been removed and the built in support for it in the ASP.NET Core pipeline is now recommended. 
`IStatusCodeHandler` has been removed.
`IModelBinder` has been removed.


JoeStead:
Nice small change then

PR #296, opened 2022-04-13
Fix GitHub Actions badge

martincostello:
Fixes the GitHub Actions badge by specifying the correct default branch.


jchannon:
Mercy buckets!!

PR #303, opened 2022-10-21
Publish the `CarterTemplate` package automatically (with MSBuild)

mderriey:
Related to #298.
Discussed in https://github.com/CarterCommunity/Carter/pull/299#issuecomment-1257956828

## Context

Please see the "Goal" and "Considerations" sections of #302.
This PR achieves the same result but with a different method, which is explained below.

## Detailed walkthrough

I kept thinking at the solution used in #302, which, while it works, brings a lot of changes and cognitive load:

- It runs MinVer as a CLI tool so we can get the computed version
- It overrides the MinVer computed version within the GitHub workflow as we're not using MSBuild anymore
- It then runs a PowerShell script to update the `csproj` file

Overall, I found it a bit too complicated.

I thought about how we could keep the process closer to what it is currently, and came to the conclusion that executing our logic (overriding the MinVer version for PRs, and updating the `csproj` file) within MSBuild would be the easiest way to do this.

### Updating the `csproj` file

MSBuild allows declaring tasks with C# code that is then compiled by Roslyn. C# code in `.targets` file is yuck, but what we're doing here is simple.

The steps are as follows:

1. We import a `targets` file from the Template project: https://github.com/CarterCommunity/Carter/pull/303/files#diff-800f45f46df5585e5ef019f729b3fd356ae505f2e3cf7870866074ba39244967R36
2. We declare our custom task which does the swap between `<ProjectReference>` and `PackageReference`: https://github.com/CarterCommunity/Carter/pull/303/files#diff-b9d07d74605b6299754f85f8bb50c521c8c6fe2ce642ede2df1dca75faa4a019R19-R43
3. We declare a target that runs our task: https://github.com/CarterCommunity/Carter/pull/303/files#diff-b9d07d74605b6299754f85f8bb50c521c8c6fe2ce642ede2df1dca75faa4a019R45-R49
4. We mark our target as a dependency to the NuGet pack task: https://github.com/CarterCommunity/Carter/pull/303/files#diff-b9d07d74605b6299754f85f8bb50c521c8c6fe2ce642ede2df1dca75faa4a019R16

## Result

When PRs get merged or the repo is tagged, we'll now publish both the `Carter` and the `CarterTemplate` packages at the same time.

For example, if the repo is tagged with `7.0.0`:

1. We'll publish the `Carter` package version 7.0.0 to NuGet, and
2. We'll publish the `CarterTemplate` package version 7.0.0; when we scaffold the project using this template, it will reference `Carter` version 7.0.0.

PR #304, opened 2022-11-04
Add workflow to automate updating references to Carter NuGet package

mderriey:
Sort of related to #298, and discussed in https://github.com/CarterCommunity/Carter/pull/299#issuecomment-1257956828

This pull request adds a new manual workflow that crawls through the `.csproj` files in the repo and updates references to the Carter NuGet package.
The workflow uses the NuGet API to find the latest published version of the package; there's the option of including prerelease versions in the search process.

### Lots of junk in the workflow file

Manual workflows can only be run if their workflow file are in the default branch, which is not the case here.
As a result, I added a `push` trigger to test out the process.
I've highlighted the portions that will be removed before this gets merged if deemed aceptable.

_Edit: that's now been removed in https://github.com/CarterCommunity/Carter/pull/304/commits/3448d3323a99624b7400808bf54369367d2786b0_

### What about the other repositories?

You mentioned that you were interested in having such a process for several repos under the Carter organisation.
This PR deliberately focuses on a single repository, and I've thought about how to use it in several repositories.
If you're happy with this and it gets merged, I'll open a new issue with my thoughts and we can take it further.

### Test runs

Here's [a run](https://github.com/CarterCommunity/Carter/actions/runs/3504732159) where we include prerelease versions, and commit and push.

We can see it worked nicely, and resulted in https://github.com/CarterCommunity/Carter/commit/b3e0e7b2527857cc5a8a2134aa321780da377d33, which looks like it was authored by Jonathan; it was of course reverted in https://github.com/CarterCommunity/Carter/commit/516dd022daca745f52d383def158bb57fc5f4482.

### Changes required before this can get merged

- [x] Remove the "push" trigger
- [x] Remove the step run during a "push" trigger
- [x] Remove the "check if workflow_dispatch" conditions since it'll become the only trigger
- [x] Uncomment the part where we commit and push

### DIFF
--- a/.github/workflows/update-carter-package-references.yml
+++ b/.github/workflows/update-carter-package-references.yml
@@ -1,9 +1,6 @@
 name: Update Carter NuGet package references
 
 on:
-  push:
-    branches:
-      - mderriey/update-carter-package-references
   workflow_dispatch:
     inputs:
       includePrerelease:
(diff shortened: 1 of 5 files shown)


jchannon:
Might be worth a chat on the Carter Slack channel about these 3 PRs to discuss what we should do going forward?

jchannon:
Do we need to merge #303 and then #304?

PR #310, opened 2022-11-23
Add possibility to specify policy names when requiring authorization at the module level

mderriey:
### Motivation

It was found by a member of the Carter community Slack that the current `RequireAuthorization` method on `CarterModule` doesn't allow to specify one or more policy names.
The result is that the only policy that is appliable at the module level is the default authorization policy configured in the app.

### Changes

This PR adds another overload of the `RequireAuthorization` that allows passing one or more policy names.

It's worth noting that ASP.NET Core has even more overloads, see https://source.dot.net/#q=requireauthorization%3CTBuilder%3E:

- `RequireAuthorization<TBuilder>(this TBuilder builder, AuthorizationPolicy policy)`
- `RequireAuthorization<TBuilder>(this TBuilder builder, Action<AuthorizationPolicyBuilder> configurePolicy)`
- `RequireAuthorization<TBuilder>(this TBuilder builder, params IAuthorizeData[] authorizeData)`

Do we want to implement some of those?

### Tests

Writing tests for this involved quite a bit of plumbing code.
If you'd rather not have tests, or if you feel there's an easier way to test this than building a server and firing a request, please let me know.

### DIFF
--- a/src/Carter/CarterExtensions.cs
+++ b/src/Carter/CarterExtensions.cs
@@ -41,14 +41,7 @@ public static class CarterExtensions
 
                 if (carterModule.requiresAuthorization)
                 {
-                    if (carterModule.authorizationPolicyNames is null)
-                    {
-                        group = group.RequireAuthorization();
-                    }
-                    else
-                    {
-                        group = group.RequireAuthorization(carterModule.authorizationPolicyNames);
-                    }
+                    group = group.RequireAuthorization(carterModule.authorizationPolicyNames);
                 }
 
                 if (!string.IsNullOrWhiteSpace(carterModule.corsPolicyName))
--- a/src/Carter/CarterExtensions.cs
+++ b/src/Carter/CarterExtensions.cs
@@ -41,7 +41,14 @@ public static class CarterExtensions
 
                 if (carterModule.requiresAuthorization)
                 {
-                    group = group.RequireAuthorization();
+                    if (carterModule.authorizationPolicyNames is null)
+                    {
+                        group = group.RequireAuthorization();
+                    }
+                    else
+                    {
+                        group = group.RequireAuthorization(carterModule.authorizationPolicyNames);
+                    }
                 }
 
                 if (!string.IsNullOrWhiteSpace(carterModule.corsPolicyName))
--- a/src/Carter/ICarterModule.cs
+++ b/src/Carter/ICarterModule.cs
@@ -33,6 +33,8 @@ public abstract class CarterModule : ICarterModule
 
     internal bool requiresAuthorization;
 
+    internal string[] authorizationPolicyNames;
+
     internal string cacheOutputPolicyName;
 
     internal readonly string basePath = "/";
@@ -64,6 +66,19 @@ public abstract class CarterModule : ICarterModule
     public CarterModule RequireAuthorization()
     {
         this.requiresAuthorization = true;
+        this.authorizationPolicyNames = null;
+        return this;
+    }
+
+    /// <summary>
+    /// Add authorization to all routes
+    /// </summary>
+    /// <param name="policyNames">A collection of policy names. If empty, the default authorization policy will be used.</param>
+    /// <returns></returns>
+    public CarterModule RequireAuthorization(params string[] policyNames)
+    {
+        this.requiresAuthorization = true;
+        this.authorizationPolicyNames = policyNames;
         return this;
     }
 
--- a/src/Carter/ICarterModule.cs
+++ b/src/Carter/ICarterModule.cs
@@ -33,7 +33,7 @@ public abstract class CarterModule : ICarterModule
 
     internal bool requiresAuthorization;
 
-    internal string[] authorizationPolicyNames;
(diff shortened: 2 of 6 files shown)


src/Carter/ICarterModule.cs
```diff
@@ -64,6 +66,19 @@ protected CarterModule(string basePath)
...
+    }
+
+    /// <summary>
+    /// Add authorization to all routes
+    /// </summary>
+    /// <param name="policyNames">A collection of policy names. If empty, the default authorization policy will be used.</param>
+    /// <returns></returns>
+    public CarterModule RequireAuthorization(params string[] policyNames)
+    {
+        this.requiresAuthorization = true;
```
mderriey, line 80: I'm not checking whether `policyNames` is `null` here.

If it is, then the module will require the default authorization policy.
I suppose it could lead to unexpected behaviors for consumers.

Let me know if you think we should add a null check.

jchannon: I think it's fine as the underlying stuff will blow if they supply null

jchannon (commented):
I'm in two minds about the tests as what we're really doing is testing the underlying auth logic. I mean we should probably have tests to see if a module specifies RequiresAuthorization() we get a 401 response maybe

src/Carter/ICarterModule.cs
mderriey: That's the thing, I don't think it will as we'll call the parameter-less overload of `RequireAuthorization` if they do.

jchannon: Actually because it's params they could just call RequiresAuthorization() with no string values so you could get rid of the method that doesn't have any args

JoeStead: You'd have to change this: https://github.com/CarterCommunity/Carter/pull/310/files#diff-ff4fe2902bfc85b0e2bba954c7b04ac67fb33dbd0c34ea6319c9ae7182dba825R44 to check for an empty array instead of null too, but that's a nicer way of doing it imo

jchannon: nope, no need i don't think as we just pass it to aspnet core. if our array is populated great, if it's empty great but we just call `group.RequireAuthorization(carterModule.authorizationPolicyNames);`

mderriey: Yep, OK.

mderriey:
> I'm in two minds about the tests as what we're really doing is testing the underlying auth logic. I mean we should probably have tests to see if a module specifies RequiresAuthorization() we get a 401 response maybe

I rewrote the tests in https://github.com/CarterCommunity/Carter/pull/310/commits/f48051a6dcb8dc445ab15eedc3cceb59e01ed801 to inspect the endpoints to make sure that the authorization-specific metadata is there.

I'm still not super happy as we're testing the internal implementation of ASP.NET Core when we call `RequireAuthorization`, but at least we don't make HTTP calls to validate the behavior.

Thoughts?

src/Carter/ICarterModule.cs
mderriey: Done in https://github.com/CarterCommunity/Carter/pull/310/commits/b0c3100ca66bd8eeaee4e28ca48ffbc5255d8395.

src/Carter/CarterExtensions.cs
```diff
@@ -41,7 +41,14 @@ public static IEndpointRouteBuilder MapCarter(this IEndpointRouteBuilder builder
 
                 if (carterModule.requiresAuthorization)
                 {
-                    group = group.RequireAuthorization();
+                    if (carterModule.authorizationPolicyNames is null)
```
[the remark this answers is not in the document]
mderriey: I believe we do, the underlying `RequireAuthorization` will throw if the array is null.

See https://source.dot.net/#Microsoft.AspNetCore.Authorization.Policy/AuthorizationEndpointConventionBuilderExtensions.cs,59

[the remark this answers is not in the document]
jchannon: Just initialize `authorizationPolicyNames` on the module as `Array.Empty<string>()` then we don't need this so if the user does:

RequiresAuthorization() = good
RequiresAuthorzation("foo") = good
RequiresAuthorzation("foo","bar") = good
RequiresAuthorzation(null) = it blows under the hood but fine, they're being dumb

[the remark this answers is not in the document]
mderriey: Ah right, I hadn't thought about this, my bad.

Addressed in https://github.com/CarterCommunity/Carter/pull/310/commits/1de8093afe85d778b1e8acd4b4530d014c46dd5e.

jchannon (approved):

cbrevik:
Not sure what the release cycle on Carter is, but could possibly this feature be released with a `7.0.1` version? 😅 

@jchannon

jchannon:
@cbrevik done! =)

cbrevik:
Wow I did not expect that quick of a feedback, thanks a bunch @jchannon! Running back from the coffee machine to continue work then I guess! 😂

jchannon:
"Get back to work!" =)

PR #324, opened 2023-07-28
Changing the scope of Validators to Scoped.

crodriguesbr:
The change is necessary to allow Validators that use external access, such as database queries, to be possible. In this case, I was using an NHibernate Session configured as Scoped. Still, I received an exception stating that my Validator is registered as Singleton, trying to access a service configured as scoped.
According to FluentValidation documentation, the default record is as Scoped (https://docs.fluentvalidation.net/en/latest/di.html)
I made the adjustments and did some tests, and everything worked usually.

### DIFF
--- a/src/Carter/CarterConfigurator.cs
+++ b/src/Carter/CarterConfigurator.cs
@@ -162,5 +162,4 @@ public class CarterConfigurator
         this.ValidatorServiceLifetime = serviceLifetime;
         return this;
     }
-
 }
--- a/src/Carter/CarterConfigurator.cs
+++ b/src/Carter/CarterConfigurator.cs
@@ -3,6 +3,7 @@ namespace Carter;
 using System;
 using System.Collections.Generic;
 using FluentValidation;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 
 /// <summary>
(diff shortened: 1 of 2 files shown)


jchannon:
I wonder if we do like FV does and pass a lifetime variable into AddCarter and set the default value to be singleton. I'm not sure of the potential impact caused by switching it to Scoped for existing users.

crodriguesbr:
I agree; I will put a parameter to define the Lifetime and leave the Singleton as default.

jchannon:
Thank will try and get this in and deployed soon

dcernach:
Could someone help with the approval of this pull request? I believe it's important for many! I was about to start a similar change and submit a pull request to address this situation; the change seems relatively straightforward. I believe this pull request could solve most of the issues with validators that use scoped services to validate requests and could also be modified to allow configuring the lifecycle of validators at another opportunity.

The ability to change the lifecycle of the registration using the DI container, in my opinion, is crucial; otherwise, it would render the use of Carter.NET in applications that are being migrated to minimal APIs unfeasible. It is quite common to have validators coupled with some sort of database access. For example, when checking the uniqueness of an email address or some other value. I understand that FluentValidation objects are quite expensive to instantiate, but in my case, and I believe in the case of others, it's a price we are willing to pay.

The automatic registration of all validators in an application as Singletons becomes a deal-breaker, even affecting previous code, as it prevents the application from starting due to errors generated during the build of the DI container. One possibility would be to disable the automatic registration of validators, along with the ability to validate the request directly at the endpoint, and leave it to the developer to manually execute the validation procedure.

jchannon:
@dcernach new nuget package has been released

PR #330, opened 2023-11-15
Update target framework to dotnet 8

scardetto:
As far as I can tell there are no breaking changes to upgrade to dotnet 8. The only issue I saw was a new deprecation warning in the mock authentication test.  I checked out the referenced URL for dotnet 8 and the code sample they provide is still using the same signature.  All of the tests are still passing.

Also, it looks like the Carter.ResponseNegotiators.Newtonsoft project references the NuGet package and not the project, so I can't update that without your assistance.  If you publish a pre-release Carter package I can update the reference and run the unit tests against the pre-release version.

Additionally, I updated the FluentValidator reference to the latest version in all projects. 

Let me know if there's anything else I can do.

### DIFF
--- a/.github/workflows/dotnetcore.yml
+++ b/.github/workflows/dotnetcore.yml
@@ -24,7 +24,7 @@ jobs:
     - name: Setup .NET Core
       uses: actions/setup-dotnet@v2
       with:
-        dotnet-version: 7.*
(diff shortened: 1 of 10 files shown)


jchannon:
Great stuff! I'll try and take a proper look at this at some point, I've just triggered the Github Action

Issue #327, opened 2023-09-10
suggest for nested work


xlogex:
[                    t.IsPublic](https://github.com/CarterCommunity/Carter/blob/82cc052b54c8a36526b6292aef506be11d4ccfcb/src/Carter/CarterExtensions.cs#L242C25-L242C25)

&& (t.IsNestedPublic || t.IsPublic)

yegorandrosov:
Found workaround
```
var catalog = new DependencyContextAssemblyCatalog();
var types = catalog.GetAssemblies().SelectMany(x => x.GetTypes());
var modules = types
            .Where(t =>
                !t.IsAbstract &&
                typeof(ICarterModule).IsAssignableFrom(t)
                && (t.IsPublic || t.IsNestedPublic)
            ).ToList();

builder.Services.AddCarter(configurator: c =>
{
    c.WithModules(modules.ToArray());
});
```

PR #342, opened 2024-05-12
Support nested modules

OnkelMato:
As a developer,
I want to use nested classes,
so that I can design my application with vertical feature slices.

For vertical slices, a static class is created which contains all concerns like handler or dtos. This includes the endpoint definition through Carter modules.

This tiny fix reads all public nested classes as well as the public classes. This fixes issue #327


jchannon:
@OnkelMato why did you close this? happy to merge

OnkelMato:
I reviewed my changes and was not satisfied with the quality. A new pull request was created.

PR #349, opened 2024-05-22
Add Roslyn Analyzer project

CollinAlpert:
As promised, I have created a Roslyn Analyzer which emits a warning when a type implementing `ICarterModule` has a constructor with parameters. Feel free to look over the test cases and see if you agree with the behavior.


test/Carter.Analyzers.Tests/CarterModuleShouldNotHaveDependenciesTests.cs
```diff
@@ -0,0 +1,323 @@
...
+    [InlineData("record struct")]
+    public Task TypeWithDependency_Diagnostic(string type)
+    {
+        var code = $$"""
+                     using Carter;
+                     using Microsoft.AspNetCore.Routing;
+
+                     {{type}} MyCarterModule : ICarterModule
+                     {
+                         internal {|#0:MyCarterModule|}(string s) {}
```
jchannon, line 22: Out of interest what is the `\#0:` stuff?

src/Carter.Analyzers/Carter.Analyzers.csproj
```diff
@@ -0,0 +1,35 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <TargetFramework>netstandard2.0</TargetFramework>
```
jchannon, line 4: should this be net8.0?

src/Carter.Analyzers/Carter.Analyzers.csproj
```diff
@@ -0,0 +1,35 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
jchannon, line 1: I'd like to ship the analyser with Carter so should Carter reference this project that is then embedded in Carter's nuget package?

jchannon (commented):
Great stuff! Thanks

test/Carter.Analyzers.Tests/CarterModuleShouldNotHaveDependenciesTests.cs
CollinAlpert: This is Roslyn's markup syntax, allowing to specify where a diagnostic is supposed to be emitted, without having to specify the exact line and column span at which the diagnostic occurs in the `DiagnosticResult`. Hence the `.WithLocation(0);`. If the analyzer emitted diagnostics at multiple locations, I would use {|#1: and then call `.WithLocation(1);`.

src/Carter.Analyzers/Carter.Analyzers.csproj
CollinAlpert: It could be, however I'd advise against it if you want maximum compatibility. Roslyn components are supposed to target netstandard2.0. I've noticed targeting higher versions works too, however not for everybody. Rider always works fine, it might just not work for some Visual Studio users. For most Visual Studio users it will also be fine. 

Considering you want to ship this with the Carter package, I say go for it and if we notice that the analyzer does not kick in for the majority of users, we can still separate this into a separate package. I will make the necessary adjustments.

CollinAlpert:
I managed to pack the analyzer with the Carter package and still keep the analyzer project `netstandard2.0`. This will ensure maximum compatibility across Visual Studio versions.

Also since the analyzer is now also added to the Carter project, we got a hit. The `CarterModule` class has a constructor with a parameter and thus the analyzer flags this. I've disabled the warning for now, however let me know if you want me to tweak the analyzer in this regard.

jchannon:
Looking good - excuse my ignorance but any reason why we can't delete the analyzer project and just put the analyzer classes inside Carter?

src/Carter/Carter.csproj
```diff
@@ -13,21 +13,27 @@
         <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
         <MinVerSkip Condition="'$(Configuration)' == 'Debug'">true</MinVerSkip>
         <PackageReadmeFile>README.md</PackageReadmeFile>
+        <EnforceExtendedAnalyzerRules>true</EnforceExtendedAnalyzerRules>
+        <NoWarn>RS2008</NoWarn>
     </PropertyGroup>
     <ItemGroup>
         <None Include="..\..\media\carterlogo.png" Pack="true" PackagePath="\" />
         <None Include="..\..\README.md" Pack="true" PackagePath="\"/>
+        <None Include="$(OutputPath)\$(AssemblyName).dll" Pack="true" PackagePath="analyzers/dotnet/cs" Visible="false" />
```
jchannon, line 22: Excuse my ignorance again - what does this do?

src/Carter/DependencyContextAssemblyCatalog.cs
```diff
@@ -89,7 +89,9 @@ private static Assembly SafeLoadAssembly(AssemblyName assemblyName)
     {
         try
         {
+#pragma warning disable RS1035
```
jchannon, line 92: Should we add this to the NoWarn csproj list?

src/Carter/ModelBinding/BindExtensions.cs
```diff
@@ -88,12 +88,14 @@ public static async Task BindAndSaveFile(this HttpRequest request, string saveLo
 
     private static async Task SaveFileInternal(IFormFile file, string saveLocation, string fileName = "")
     {
+#pragma warning disable RS1035
```
jchannon, line 91: Should we add this to the NoWarn csproj list?

src/Carter/Carter.csproj
CollinAlpert: Please don't hold back with questions – it's important you understand the code I added :)

To answer your question: It's not enough to add the code to the project. To make analyzers work inside a NuGet package, they must be placed in a specific directory inside the package, in this case the `analyzers/dotnet/cs` directory, since it is a C# analyzer.

src/Carter/DependencyContextAssemblyCatalog.cs
CollinAlpert: No, since this warning would be valid if `DependencyContextAssemblyCatalog` were an analyzer. Since it is not, the warning can be suppressed. Just one small caveat of not splitting up library code and analyzers.

CollinAlpert:
Yes, you will, however the project reference must look like this to include analyzer behavior:
`<ProjectReference Include="../../src/Carter/Carter.csproj" OutputItemType="Analyzer" />`

`OutputItemType` is not necessary for `PackageReference` elements, since we pack the DLL in the analyzer directory, something that is not possible to `ProjectReference` elements.

jchannon:
Should rename it to `Carter` from `CARTER1`?
Screenshot 2024-05-24 at 12 06 37

CollinAlpert:
We could, however I wouldn't recommend it. The [guidelines for diagnostic IDs](https://learn.microsoft.com/en-us/dotnet/csharp/roslyn-sdk/choosing-diagnostic-ids) recommend using the project name + a number. If you decide you want another analyzer to warn about something else in the future, you'd need to distinguish it from this one, e.g. by naming it `CARTER2`. The compiler diagnostics (`CSXXXX`), code analysis diagnostics (`CAXXXX`) and system library diagnostics (`SYSLIBXXXX`) all follow this pattern.

jchannon:
Can you rebase, I've added some permissions to the actions workflow file :)

jchannon:
Actually GH has a button to do it :)

jchannon:
Thanks for this!! Appreciate it 👍 👍

CollinAlpert:
Happy to be of assistance! Ping me if you want any other rules implemented 😄

PR #350, opened 2024-05-24
Support for nested modules

OnkelMato:
As a developer,
I want to use nested classes,
so that I can design my application with vertical feature slices.

For vertical slices, a static class is created which contains all concerns like handler or dtos. This includes the endpoint definition through Carter modules.

This tiny fix reads all public nested classes as well as the public classes. This fixes issue https://github.com/CarterCommunity/Carter/issues/327


jchannon (approved):

jchannon:
Thanks! 👍

PR #378, opened 2025-08-16
Make validator lifetimes configurable on a validator-by-validator level

enkelm:
While on a work related project, I came across a DI runtime exception for 'Scoped' lifetime Validators. After doing some digging, I came across [this issue](https://github.com/CarterCommunity/Carter/issues/123) and then [this PR](https://github.com/CarterCommunity/Carter/pull/324) which sets a global lifetime for all validators. This PR aims to add a method to set up custom logic based on type inference on a validator-by-validator level.

### Implementation TLDR:
- Introduces `ValidatorServiceLifetimeFactory` of type `Func<Type, ServiceLifetime>` on `CarterConfigurator`.
- Default factory:
  - Checks for a `[ValidatorLifetime]` attribute on the validator type → uses its `Lifetime`.
  - If no attribute → falls back to `ValidatorServiceLifetime` (global default).
- Users can override the factory via `WithValidatorServiceLifetimeFactory(...)` at startup.
- `CarterExtensions` updated to register validators with lifetimes resolved per type.
- `IValidatorLocator` lifetime is set to the lowest of all registered validator lifetimes.

### Step by step:
On this first draft, it's implemented by the `ValidatorServiceLifetimeFactory` field on `CarterConfigurator` of type `Func<Type, ServiceLifetime>`. This func takes the validator type as an argument and must return a `ServiceLifetime`. It can be passed to the config on startup with the `WithValidatorServiceLifetimeFactory` but it also has a default implementation. By default, it checks for the newly added `ValidatorLifetimeAttribute` on the validator's type and return its `Lifetime` property or the config's `ValidatorServiceLifetime` if there isn't any attribute found.

This all comes together on `CarterExtensions` where the `WireupCarter` validator DI logic is changed to set the lifetime of each validator based on the result of  `ValidatorServiceLifetimeFactory`. As for the `IValidatorLocator` lifetime, that's set to the lowest existing lifetime.

This is by no means ready to merge but more like a POC. I have tested it on my own app and seems to be working so far.

**Note:** Changing `ValidatorServiceLifetime` -> `DefaultValidatorServiceLifetime` is done for readability more than anything. I don't think it should go through since it breaksuser space.

### DIFF
--- a/src/Carter/CarterConfigurator.cs
+++ b/src/Carter/CarterConfigurator.cs
@@ -2,6 +2,7 @@ namespace Carter;
 
 using System;
 using System.Collections.Generic;
+using System.Reflection;
 using FluentValidation;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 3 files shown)


jchannon:
Might be worth adding an example in the Carter.Samples app to show how you might use this new feature

PR #381, opened 2025-11-29
fix: enhance module type visibility checks in CarterExtensions

vinhdev17:


vinhdev17:
> Quote reply

I just added some tests 👍

jchannon:
Can you change `ResponseExtensions` to have this `negotiator = negotiators.First(x => x.GetType() == typeof(DefaultJsonResponseNegotiator));` when `negotiator` is null as it's pulling in the InternalResponseNegotiator you added

vinhdev17:
image

From what I observed, all negotiators are being resolved, including the test ones. To handle this scenario, I introduced new attributes to filter them out accordingly.
Let me know if this approach aligns with what you had in mind.

jchannon (approved):

